PR #6, opened 2020-04-18
Updated the build system, added a NOMMU target

Darchiv:


Makefile
```diff
@@ -85,7 +34,7 @@ DRIVERS_SRCS = netif-driver.c
 DRIVERS_SRCS_UTIL = bdring.c pktmem.c physmmap.c res-create.c
 DRIVERS_SRCS_PPPOS = pppos.c
 DRIVERS_SRCS_TUNTAP = tuntap.c
-include Makefile.$(TARGET)
+include _targets/Makefile.$(TARGET_FAMILY)-$(TARGET_SUBFAMILY)
```
nalajcie, line 37: add "-" to allow no per-target makefile present

Makefile
```diff
@@ -85,7 +34,7 @@ DRIVERS_SRCS = netif-driver.c
 DRIVERS_SRCS_UTIL = bdring.c pktmem.c physmmap.c res-create.c
```
nalajcie, line 34: Not directly related to this patch, but it should be `:=`  instead of `=` here (this applies in most other places also). See https://www.gnu.org/software/make/manual/make.html#Reading-Makefiles for details.

Makefile
Darchiv: How about doing just `NET_DRIVERS ?= pppos` in target-specific Makefile and then `DRIVERS_SRCS += $(foreach driver,$(NET_DRIVERS), $(DRIVERS_SRCS_$(driver)))` here? Target-specific sources (like `DRIVERS_SRCS_enet`) can be defined only in target-specific Makefile and it will still work.

nalajcie: yeah, I like this idea, much better than directly specifying C files to compile. Should we add explicit check if particular driver exists to prevent typos? (`NET_DRIVERS = ppos` should fail or just compile nothing?)

Darchiv: I have added such a check in the update. Every target Makefile defines a list of supported drivers so that a driver unsupported on this target will not be silently skipped, compared to a global list.

Targets without a Makefile will still work (they just need extra `NET_DRIVERS_SUPPORTED` in env).

Makefile
Darchiv: Fixed in the update.

Makefile
nalajcie: ok, maybe more straightforward would be to check if `DRIVERS_SRCS_ppos` is non-empty, but Your solution is also fine by me. I'm ok with merging this right away.

PR #7, opened 2020-04-29
Ported to NOMMU, added support for another modem

Darchiv:
Destination branch TBD

### DIFF
--- a/drivers/pppos.c
+++ b/drivers/pppos.c
@@ -25,7 +25,10 @@
 #include <sys/msg.h>
 #include <syslog.h>
 
+#include <pppos_modem.h>
+
 enum {
+	CONN_STATE_DISCONNECTING,
 	CONN_STATE_DISCONNECTED,
 	CONN_STATE_CONNECTING,
 	CONN_STATE_CONNECTED,
@@ -38,10 +41,13 @@ typedef struct
 
 	const char *serialdev_fn;
 	const char *serialat_fn;
+#if PPPOS_USE_CONFIG_FILE
 	const char *config_path;
 	char apn[64];
+#endif
 	int fd;
 
+	volatile int want_connected;
 	volatile int conn_state;
 	handle_t lock, cond;
 
@@ -177,7 +183,6 @@ static int at_check_result(const char* buf)
 
 	while (at_result_codes[res]) {
 		if ((result = strstr(buf, at_result_codes[res])) != NULL) {
-			*result = '\0';
 			return res;
 		}
 		res += 1;
@@ -225,7 +230,7 @@ static int at_send_cmd_res(int fd, const char* cmd, int timeout_ms, char *rx_buf
 		} else {
 			rx_buf[off + len + 1] = '\0';
 
-			int res = at_check_result(rx_buf + off);
+			int res = at_check_result(rx_buf);
 			off += len;
 
 			if (res >= 0) {
@@ -279,6 +284,21 @@ static int at_disconnect(int fd) {
 	return res;
 }
 
+
+static int at_is_responding(int fd, int timeout_ms)
+{
+	int res;
+	int retry = 10;
+
+	while ((res = at_send_cmd(fd, "AT\r\n", timeout_ms * 1000)) != AT_RESULT_OK && retry--);
+
+	if (res != AT_RESULT_OK) {
+		log_warn("modem not responding, res=%d", res);
+		return 0;
+	}
+
+	return 1;
+}
 
 
 /****** PPPoS support functions ******/
@@ -399,7 +419,8 @@ static void pppos_do_rx(pppos_priv_t* state)
 	int len;
 	u8_t buffer[1024];
 
-	while (state->conn_state != CONN_STATE_DISCONNECTED) {
+	while (state->conn_state != CONN_STATE_DISCONNECTED
+			&& state->conn_state != CONN_STATE_DISCONNECTING) {
 		len = read(state->fd, buffer, sizeof(buffer));
 		if (len > 0) {
 			/* Pass received raw characters from PPPoS to be decoded through lwIP
@@ -422,16 +443,6 @@ static void pppos_do_rx(pppos_priv_t* state)
 }
 
 
-const char* at_init_cmds[] = {
-	"ATZ\r\n",			// reset MODEM
-	"ATQ0 V1 E0 S0=0 &C1 &D2 +FCLASS=0\r\n",	// setup serial/message exchange
-	"AT+WS46=29\r\n",		// disable LTE
-	"AT+CREG?\r\n", 		// check network registration (just for debug)
-	"AT+COPS?\r\n", 		// check operator registration (just for debug)
-	"AT+CSQ\r\n",       // check signal quality (for debug)
-	NULL,
-};
-
 static void pppos_mainLoop(void* _state)
 {
 	pppos_priv_t* state = (pppos_priv_t*) _state;
@@ -440,6 +451,12 @@ static void pppos_mainLoop(void* _state)
 	int running = 1;
 
 	while (running) {
+		mutexLock(state->lock);
+		while (!state->want_connected) {
+			condWait(state->cond, state->lock, 0);
+		}
+		mutexUnlock(state->lock);
+
 		/* Wait for the serial device */
 		if (state->fd < 0)  {
 			while ((state->fd = open(state->serialdev_fn, O_RDWR | O_NOCTTY | O_NONBLOCK)) < 0)
@@ -449,17 +466,25 @@ static void pppos_mainLoop(void* _state)
 		}
 
 		serial_set_non_blocking(state->fd);
+#if PPPOS_DISCONNECT_ON_INIT
 		if (at_disconnect(state->fd) != AT_RESULT_OK)
 			goto fail;
+#endif
 
+#if PPPOS_USE_CONFIG_FILE
 		mutexLock(state->lock);
 		while (!state->apn[0])
 			condWait(state->cond, state->lock, 0);
 		mutexUnlock(state->lock);
+#endif
+
+		if (!at_is_responding(state->fd, 1000)) {
+			goto fail;
+		}
 
 		const char** at_cmd = at_init_cmds;
 		while (*at_cmd) {
-			if ((res = at_send_cmd(state->fd, *at_cmd, 3000)) != AT_RESULT_OK) {
+			if ((res = at_send_cmd(state->fd, *at_cmd, AT_INIT_CMDS_TIMEOUT_MS)) != AT_RESULT_OK) {
 				log_warn("failed to initialize modem (cmd=%s), res=%d, retrying", *at_cmd, res);
 				goto fail;
 			}
@@ -467,6 +492,7 @@ static void pppos_mainLoop(void* _state)
 			at_cmd += 1;
 		}
 
+#if PPPOS_USE_CONFIG_FILE
 		{ /* Configure APN */
 			char at_set_apn[256];
 			if (snprintf(at_set_apn, sizeof(at_set_apn), "AT+CGDCONT=1,\"IP\",\"%s\"\r\n", state->apn) >= sizeof(at_set_apn)) {
@@ -479,14 +505,12 @@ static void pppos_mainLoop(void* _state)
 				goto fail;
 			}
 		}
+#endif
 
-		if ((res = at_send_cmd(state->fd, "AT+CGDATA=\"PPP\",1\r\n", 3000)) != AT_RESULT_CONNECT) {
-			log_warn("failed to dial PPP, res=%d, retrying", *at_cmd, res);
+		if ((res = at_send_cmd(state->fd, AT_CONNECT_CMD, AT_CONNECT_CMD_TIMEOUT_MS)) != AT_RESULT_CONNECT) {
+			log_warn("failed to dial PPP, res=%d, retrying", res);
 			goto fail;
 		}
-
-		// TODO: provide authentication params externally
-		//ppp_set_auth(state->ppp, PPPAUTHTYPE_PAP, "plusgsm", "plusgsm");
 
 		log_debug("ppp_connect");
 		state->conn_state = CONN_STATE_CONNECTING;
@@ -507,8 +531,12 @@ static void pppos_mainLoop(void* _state)
 		}
 		mutexUnlock(state->lock);
 
+		log_debug("connection closed");
+
 fail:
-		log_error("connection has failed, retrying");
+		serial_close(state->fd);
+		state->fd = -1;
+
 		sleep(PPPOS_CONNECT_RETRY_SEC);
 	}
 
@@ -524,6 +552,7 @@ fail:
 
 static int pppos_netifUp(pppos_priv_t *state)
 {
+#if PPPOS_USE_CONFIG_FILE
 	char lcfg[256] = { 0 };
 	int line = 0;
 	FILE *fcfg = fopen(state->config_path, "r");
@@ -565,9 +594,13 @@ static int pppos_netifUp(pppos_priv_t *state)
 			log_warn("[line %d] unsupported option: %s (val: %s)", line, lcfg, cfgval);
 	}
 
+	fclose(fcfg);
+#else
+	mutexLock(state->lock);
+#endif
+	state->want_connected = 1;
 	condSignal(state->cond);
 	mutexUnlock(state->lock);
-	fclose(fcfg);
 
 	return 0;
 }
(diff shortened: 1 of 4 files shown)


port/main.c
```diff
@@ -132,17 +132,17 @@ static void mainLoop(void)
 	status_oid.port = port;
 	pf_oid.port = port;
 
-	if (create_dev(&route_oid, "/dev/route") < 0) {
+	if (portRegister(port, "/dev/route", &route_oid) < 0) {
```
agkaminski, line 135: This change will make MMU target regress. create_dev is not working because NOMMU lacking filesystem which can handle mkdir("/dev", 0666);

We can go three ways here:
- make distinction in lwip (so use portRegister on xxx target, use create_dev on yyy target)
- handle this in libphoenix (use portRegister when mkdir failed)
- ensure that every system have some sort of mkdir capable rootfs.

@nalajcie Which way you think is best? I'm leaning to handling in libphoenix option, personally.

nalajcie: I think `create_dev` is part of our standard library API so it should work on every supported platform so I would opt for always using `create_dev` and fixing the implementation in libphoenix.

To be on the same page: `mkdir` is a fallback in case `devfs` lookup fails (see https://github.com/phoenix-rtos/libphoenix/blob/6fb4fa6b3f582931ad5a336bba004cf91069e951/unistd/file.c#L219) - maybe we can provide some simple NOMMU devfs instead (e.g. without directory support?). I'm not familiar with phoenix-rtos convention - does it have to be a separate process or can it be handled differently (in-kernel? ;))?

As You suggested - the quick fix is to use `portRegister` in `create_dev` implementation - maybe we should always do it on NOMMU platforms for consistency? I don't know all of the drawbacks of this solution so I can't recommend the specific way.

nalajcie:
Update: I've read the PR description that target brach for this PR's should not be master, so MMU targets won't be affected.

Even so, I would prefer to strip out AT commands ("chat script") from PPPoS implementation and keep it modem-agnostic, but my understanding is that we don't have real filesystem for NOMMU so keeping AT commands in a file is not a solution we can implement right away.

Maybe we should try to keep changes between this branch and the master minimal (eg. change pppos logic on master branch, ifdef out instead of deleting the code reading configuration file (as not working right now)? This would simplify the future merge when it would be possible.

@agkaminski What do you think? Should we leave it in a branch as-is or split this into changes for master and modem-specific branch? Or maybe You have some other ideas regarding passing the AT commands to lwip?

agkaminski:
For now we can keep this branch as is, but we our goal is to merge all changes into master.
As for AT commands, I don't see a point in keeping them in a file. It can be compiled into lwip as well.

Darchiv:
The `portRegister()` were reverted back to `create_dev()`. I suggest a _nommu-modem_ branch for now, until the AT commands are separated/taken care of properly and the PPPoS is mergeable.

drivers/pppos.c
```diff
@@ -440,6 +463,12 @@ static void pppos_mainLoop(void* _state)
 	int running = 1;
 
 	while (running) {
+		mutexLock(state->lock);
+		while (!(state->ppp->netif->flags & NETIF_FLAG_UP)) {
```
nalajcie, line 467: **This works just by accident.**

The `netif->flags` is not a volatile field, so compiler may optimize it out (keep the value in register).

We should probably add `volatile int want_connected;` to `state` struct and make decision based on that.

side note: `netif` is available directly in `state`.

drivers/pppos.c
```diff
@@ -524,61 +534,19 @@ static void pppos_mainLoop(void* _state)
...
 }
 
 
 static int pppos_netifDown(pppos_priv_t *state)
 {
-	if (state->apn[0]) {
-		state->apn[0] = 0;
-		pppapi_close(state->ppp, 0);
-	}
+	/* Unconditional use of pppapi_close() in the status callback
```
nalajcie, line 545: good catch, I see some Kamil added this code. I suppose Your solution should break the connection more gracefully. 

This will work across the threads as conn_state is declared volatile. It can also be done with `want_connected` field I suggested above.

drivers/pppos.c
```diff
@@ -448,46 +477,21 @@ static void pppos_mainLoop(void* _state)
...
-			if (snprintf(at_set_apn, sizeof(at_set_apn), "AT+CGDCONT=1,\"IP\",\"%s\"\r\n", state->apn) >= sizeof(at_set_apn)) {
-				log_error("APN name too long");
-				goto fail;
-			}
-
-			if ((res = at_send_cmd(state->fd, at_set_apn, 3000)) != AT_RESULT_OK) {
-				log_warn("failed to set APN, retrying");
-				goto fail;
-			}
+		if ((res = at_send_cmd(state->fd, "AT+CGATT=0\r\n", 3000)) != AT_RESULT_OK) {
```
anglov, line 485: I wonder if this timeout is not to small. When I read different modem documentation and chat script examples  I checked that modems requires wait from 10 to 300 seconds. If i read correctly we are waiting only 3 seconds.
What do you think @nalajcie ?

nalajcie: I'm not sure why this command is needed (I don't know the exact modem version either, so I can't check it in datasheet) - maybe without it the `AT+CGDATA` fails?

This command usually is used to disconnect GPRS context, so it should be pretty quick.

If You meant the next command (`AT+CGDATA`) I think it should return `CONNECT` immediately and start connecting (next states will be available via PPP callback), but it might be up to manufacturer to mess this up :).

For LTE modems it should be pretty quick, as the primary PDP context is received and attached during network registration (NB-IoT may be similar).

I don't see any command for checking if we're even registered in the network (previously it was not enforced/waited for, just outputted to the log, as we're in an infinite loop).

drivers/pppos.c
```diff
@@ -280,6 +280,28 @@ static int at_disconnect(int fd) {
 }
 
 
+static int at_setup(pppos_priv_t* state)
+{
+	int res;
+	int retry = 10;
+
+	usleep (1 * 1000 * 1000);
```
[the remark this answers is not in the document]
Darchiv: Done, removed it from there.

drivers/pppos.c
```diff
@@ -280,6 +280,28 @@ static int at_disconnect(int fd) {
...
+
+	while ((res = at_send_cmd(state->fd, "AT\r\n", 3000)) != AT_RESULT_OK && retry--);
```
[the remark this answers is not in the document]
Darchiv: I had left only the `AT` command so this function was renamed to `at_is_responding`.

drivers/pppos.c
```diff
@@ -448,46 +477,21 @@ static void pppos_mainLoop(void* _state)
 			log_info("open success!");
 		}
 
-		serial_set_non_blocking(state->fd);
```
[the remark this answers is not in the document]
Darchiv: In order to accomodate the different modems, I have put a few defines here and there. A `PPPOS_MODEM_TYPE` define is now required and it will break existing setups, but the flavor has to be somehow explicitly chosen anyway, be it in a build file or some other.

New defines now make it possible to merge the `lte430` branch with minimal work required.

drivers/pppos.c
```diff
@@ -448,46 +477,21 @@ static void pppos_mainLoop(void* _state)
...
-		if (at_disconnect(state->fd) != AT_RESULT_OK)
+		if (at_setup(state)) {
```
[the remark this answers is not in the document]
Darchiv: I have revoked these changes to minimize the divergence. When it comes to function return values, `at_is_responding` should be pretty intuitive (1 - responding, 0 - not responding).

drivers/pppos.c
```diff
@@ -448,46 +477,21 @@ static void pppos_mainLoop(void* _state)
...
+			serial_close(state->fd);
```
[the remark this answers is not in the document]
Darchiv: Revoked the changes.

drivers/pppos.c
```diff
@@ -448,46 +477,21 @@ static void pppos_mainLoop(void* _state)
...
-		while (*at_cmd) {
-			if ((res = at_send_cmd(state->fd, *at_cmd, 3000)) != AT_RESULT_OK) {
-				log_warn("failed to initialize modem (cmd=%s), res=%d, retrying", *at_cmd, res);
-				goto fail;
-			}
-
-			at_cmd += 1;
 		}
 
-		{ /* Configure APN */
```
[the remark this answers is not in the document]
Darchiv: The `PPPOS_USE_CONFIG_FILE` feature flag now guards the APN configuration.

drivers/pppos.c
```diff
@@ -448,46 +477,21 @@ static void pppos_mainLoop(void* _state)
...
+			serial_close(state->fd);
 			goto fail;
-
-		mutexLock(state->lock);
-		while (!state->apn[0])
-			condWait(state->cond, state->lock, 0);
-		mutexUnlock(state->lock);
-
-		const char** at_cmd = at_init_cmds;
-		while (*at_cmd) {
```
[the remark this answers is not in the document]
Darchiv: Done, `at_init_cmds` has come back.

drivers/pppos.c
```diff
@@ -524,61 +534,19 @@ static void pppos_mainLoop(void* _state)
...
-			cfgval[strlen(cfgval) - 1] = 0;
-		}
-
-		if (!strcasecmp(lcfg, "apn"))
-			strncpy(state->apn, cfgval, sizeof(state->apn) - 1);
-		else
-			log_warn("[line %d] unsupported option: %s (val: %s)", line, lcfg, cfgval);
-	}
-
 	condSignal(state->cond);
```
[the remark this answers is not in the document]
Darchiv: Done, now under the mutex. A volatile variable was added, see the comment below.

drivers/pppos.c
Darchiv: Right, I have added the volatile member. The problem should be solved now.

drivers/pppos.c
```diff
@@ -524,61 +534,19 @@ static void pppos_mainLoop(void* _state)
 
 static int pppos_netifUp(pppos_priv_t *state)
 {
-	char lcfg[256] = { 0 };
```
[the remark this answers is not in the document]
Darchiv: Done, now under the `PPPOS_USE_CONFIG_FILE` feature flag.

drivers/pppos.c
```diff
@@ -280,6 +280,28 @@ static int at_disconnect(int fd) {
...
+	usleep (1 * 1000 * 1000);
+
+	while ((res = at_send_cmd(state->fd, "AT\r\n", 3000)) != AT_RESULT_OK && retry--);
+
+	if (res != AT_RESULT_OK) {
+		log_warn("failed to reset modem, res=%d", res);
+		return -1;
+	}
+
+	if ((res = at_send_cmd(state->fd, "AT+QSCLK=0\r\n", 3000)) != AT_RESULT_OK) {
```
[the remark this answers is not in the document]
Darchiv: Removed this particular command. See the comments below.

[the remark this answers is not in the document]
Darchiv: Done, moved.

drivers/pppos.c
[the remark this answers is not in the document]
Darchiv: The modem utility has already been resetting the modem (via power up/enable pins) on every connection. I have added `ATZ` to `at_init_cmds` anyway in case there is a change that will make the serial device not reset the modem in hardware.

drivers/pppos.c
```diff
@@ -507,8 +511,14 @@ static void pppos_mainLoop(void* _state)
 		}
 		mutexUnlock(state->lock);
 
+		log_debug("connection closed");
+
 fail:
-		log_error("connection has failed, retrying");
+		// if (fcntl(state->fd, F_GETFD) < 0) {
```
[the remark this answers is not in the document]
Darchiv: Done, removed.

drivers/pppos.c
Darchiv: The reason for inclusion of `AT+CGATT=0` was that the modem had been attaching to a PDP context automatically and in result did not allow PPP. I have replaced `AT+CGATT=0` with `AT+QCFG="autopdn",0` and now it seems the connection is established faster. The debug commands (like `AT+CREG?`) are back.

Darchiv:
I have addressed all the issues and merging different flavors/modems should be easier now.

drivers/pppos.c
```diff
@@ -25,7 +25,52 @@
...
+#elif PPPOS_MODEM_TYPE == PPPOS_MODEM_QUECTEL
+
+#define PPPOS_USE_CONFIG_FILE 0
+#define PPPOS_DISCONNECT_ON_INIT 0
+#define AT_CONNECT_CMD "AT+CGDATA=\"PPP\",1\r\n"
+#define AT_INIT_CMDS_TIMEOUT_MS 3000
+#define AT_CONNECT_CMD_TIMEOUT_MS 3000
+
+static const char* at_init_cmds[] = {
+	"ATZ\r\n",
```
nalajcie, line 59: it would be great to add some comments what these AT commands are doing without looking at the documentation.

Note: missing ECHO off command we were discussing previously (this would clean up the logs a little)

drivers/pppos.c
```diff
@@ -25,7 +25,52 @@
 #include <sys/msg.h>
 #include <syslog.h>
 
+/* Intentionally does not begin from 1. */
+#define PPPOS_MODEM_GENERIC 2
```
nalajcie, line 29: Note: this was actually meant for a Telit LTE modem, not a generic one, but I don't know if we want to specify it that much (it may work with other modems also).

Note for the future (I'm okay with the current approach, but it's not extensible in regards of adding many more modems):
I would prefer to put the configuration in separate headers (one modem per header, eg. `pppos_modem.h` in subdirs eg. `modem/[modem_name]/pppos_modem.h`) and just decide on the modem in Makefile (add correct include path to `CFLAGS`). This would make adding new modems less painful.

@anglov, do You think these per-modem defines are enough to provide support for LTE cat M1 modem You were experimenting with without too many additional changes?

anglov: In cat M1 modem there was one issue: when modem go into data mode, they doesn't back to AT mode after '+++'. Maybe just some additional configuration was needed but I didn't find solution for that.

In fact I just need to disable code for setting APN (chosing the default one), disablr action that needs to back to AT mode and change init commands.

So, it's possible that different at_disconnect will be needed for cat M1, but not sure - I don't find proper solution  for that modem.

What is missing and could be problematic in future is set non-default APN if configuration file is not used. Probably not an issue for this PR.

drivers/pppos.c
Darchiv: I have included the comments and also the ATE0 command (at least for Quectel).

drivers/pppos.c
Darchiv: Maybe until it is known to be generic (tested), let's call it `telit`. Regarding separate headers, at first I thought it would unnecessarily spawn many tiny files. I have separated them anyway and it looks much more manageable now - only a new subdirectory and a header are needed to support a new modem (no changes to phoenix-rtos-lwip needed), and a `PPPOS_MODEM` define set appropriately in a project.

Darchiv:
When it comes to other projects which use the `pppos` driver, they must now define `PPPOS_MODEM`, so this is a potential break point. Should there be a default (?=) modem define in Makefile, no problems will occur, but I am not sure about this solution.

drivers/pppos.c
nalajcie: Thanks, for telit it's done in one multi-command specifying serial command specifics (`E0` part is "echo off").

drivers/pppos.c
nalajcie: I like the current version :). I think there should be no default - If You want to use PPPoS driver in the project, You should also explicitly specify the modem You would like to support.

nalajcie:
> Looks good. One thing though. condSignal under mutex will cause extra reschedule (as condSignal itself causes waiting thread to be scheduled. This thread will then try to grab mutex and it will fail and we will have to wait for waker to release the mutex).
> 
> Just a notice, maybe this is wrong behaviour of condSignal.

Huh? This is kind-of unusual, eg. in the model of concurrent programming it's usually stated that the signaling should occur under the mutex (https://en.wikipedia.org/wiki/Monitor_(synchronization)#Condition_variables).

This holds true for POSIX threads also (https://pubs.opengroup.org/onlinepubs/007908799/xsh/pthread_cond_signal.html), the actual wording is:
_The `pthread_cond_signal()` or `pthread_cond_broadcast()` functions may be called by a thread whether or not it currently owns the mutex that threads calling `pthread_cond_wait()` or `pthread_cond_timedwait()` have associated with the condition variable during their waits; however, if predictable scheduling behaviour is required, then that mutex is locked by the thread calling `pthread_cond_signal()` or `pthread_cond_broadcast()`._

So we should assume the common use case is to call `condSignal` under mutex, so this code path may need to be optimized.

This is not relevant to this PR, we can put it as an issue in other repo (kernel?).

PR #8, opened 2020-08-24
Lwipopts

anglov:
First commit in set: Change ifdef to if – it allows to properly disabling some opitions in lwipopts.h. Also added some preprocessor ifs, where it missing.

Second commit – add possibility to pass lwipopts externally, depending on project. Not sure if my way how to do it is ok, so please take a look at this.


agkaminski:
This is quite fragile solution. If someone in the future change order of -I's we'll have a silent problem. But I don't have any better idea tbh

anglov:
I don't like it either. Also don't like that I need to pass whole dir with lwipopts.h file, but currently I don't have better idea how to do it. What do you think @nalajcie ?

nalajcie:
Regarding changing the `#ifdef` to `#if` -> I strongly advise to add `-Wundef` to CFLAGS (undefined macros evaluate to `0` in `#if`'s which makes nice silent error to be searched for hours :))

Regarding the include - I don't know what's the intended use-case -> just to change 1-2 values or provide a full separate `lwipopts.h`?

There are multiple possibilities to do it:
- we could use `LWIP_FEATURE_*` macros in our central `lwipopts.h` and pass them by CFLAGS (eg. `LWIP_FEATURE_PPOS`) - the :heavy_plus_sign: is that we have all available/supported configurations in a single place, :heavy_minus_sign: is that the amount of feature flags might grow over time
- (as a variation above) we could `#ifdef`-guard all defines in `lwipopts.h` and provide overrides as per-project CFLAGS (:heavy_minus_sign: might be quite messy and challenging to maintain over time)
- separate `lwipopts.h` -> introduced via `-I` in CFLAGS (as You did), subideas:
  - move current `lwipopts.h` to new directory, add it via `-I` only if LWIPOPTS_DIR is empty (actually, define `LWIPOPTS_DIR ?= "include/default-opts"`) - (:heavy_plus_sign: no way to include invalid `lwipopts.h` then; :heavy_minus_sign: all options need to be defined per-project - always copying the whole file)
  - make custom `lwipopts.h` an overlay (`#ifdef`-guard defaults and use `#include_next` in overlay; :heavy_minus_sign: we might miss that we didn't include the custom lwipopts)
  - (sum of the 2 above - the best subidea) `#include <default-ops/lwipopts.h>` in the customization so we're sure we used the customization
- support `lwipopts_custom.h` as an overlay over default opts with `__has_include` compiler extension (supported by GCC and CLANG), eg.:
```c
#if defined __has_include
#  if __has_include (<lwipopts_custom.h>)
#    include <lwipopts_custom.h>
#  endif
#else
#  warning "__has_include not supported!"
#endif
```
(:heavy_plus_sign: we may override only the options we would like to change; :heavy_minus_sign: using compiler extension, but I think it would never be a problem in the future)

Personally, I would either go with `LWIPOPTS_DIR ?= "include/default-opts"` (as an overlay with including default ops in a subdir), `LWIP_FEATURE_*` (if the intended use-case is really to enable/disable some features) or `__has_include` (I like it because it's hackish (:+1:) and the file names are more descriptive).

anglov:
About -Wundef to CFLAGS – agree
About your proposition – I like it. 
I decided to use LWIPOPTS_DIR ?= "include/default-opts" option. 

Also – to not to overwrite whole file, we need to put all defines under ifndef, to not generate redefinition warning.

I think that we need to define which opts is required to set in this file and which could be defaulted by lib-lwip/src/include/lwip/opt.h and redefined by the project.

For now I hope whole opts rewrite would be enough.

nalajcie:
> About -Wundef to CFLAGS – agree
> About your proposition – I like it.
> I decided to use LWIPOPTS_DIR ?= "include/default-opts" option.

:+1: 

> Also – to not to overwrite whole file, we need to put all defines under ifndef, to not generate redefinition warning.

Yep, that was my intent. Maybe we don't need to do it right away, just when we would need to redefine some options in it project, than we would `#ifndef`-guard it.

> I think that we need to define which opts is required to set in this file and which could be defaulted by lib-lwip/src/include/lwip/opt.h and redefined by the project.
> 
> For now I hope whole opts rewrite would be enough.

Yeah, I'm assuming all of the options should be overridable - no need to limit ourselfs. Maybe in the future we would provide some basic "profiles" (different `lwipopts.h` in this repo, not in the project repo) if many values will be shared between the devices (eg. "lowmem" for device with limited RAM (providing smaller memory pools), etc.).

But this is an idea for the future :).

Regarding "-Wundef" - I would try to add it globally (in `phoenix-rtos-build`), but that's out of scope of this PR.

nalajcie (approved):

PR #9, opened 2020-08-27
Change collector stack size

anglov:
Stack size of collector thread should be parametrize against different platform and should be modified by opts.

### DIFF
--- a/port/threads.c
+++ b/port/threads.c
@@ -30,7 +30,7 @@ typedef struct {
 
 
 static struct {
-	char collector_stack[32 * 4096] __attribute__((aligned(8)));
+	char collector_stack[512] __attribute__((aligned(8)));
 	rbtree_t threads;
 	handle_t lock;
 } global;


agkaminski (changes requested):
We need to find out why magic number is 32 and change it in all places at the same time. Also, maybe we can change 4096 part to something smaller?

nalajcie:
If You want to parametrize it, then maybe guard it with `#ifndef`.

Side note: why the collector stack size has to be `128 kB`? Are RB tree operations so stack-consuming? Maybe it's worth to investigate that.

agkaminski:
@nalajcie 4096 comes from _PAGE_SIZE on cortex-a. Lazy programming

nalajcie:
> @nalajcie 4096 comes from _PAGE_SIZE on cortex-a. Lazy programming

I'm aware of this, just don't know why it can't be for example 512 bytes. I don't see any big on-stack variables in thread main function - and it mostly calls RB manipulation functions (which I'm not familiar with - thus my questions).

What I'm seeing is commit 4a74318 and 954cc7f which suggests it really needs to be large enough but it's not evident from the code why.

anglov:
After looking at code it seems that this thread stack size is relic of other changes

nalajcie:
Are You sure?

Did You check the file at the commit I've pointed to (increasing stack size from 16kB to 128kB) - https://github.com/anglov/phoenix-rtos-lwip/blob/4a743183c587730ce90d32c78b2b7c245824f8aa/port/threads.c

The function body is the same - so unless `threadJoin()` or RB trees implementation were changed - there seems to be a valid reason for the stack to be so large (yet unknown to me :)).

Are we even able to run lwIP on i.MX6ULL to verify it won't break under heavy load?

anglov:
Well, I checked the code. All rb function is iterative (no recursion), no big structure on the stack, I don't see the reason why it couldn't work in heavy load.

I ask @Maxez to check it on intel.

nalajcie:
Hmm, maybe @js6i would be able to tell us why such a big stack was needed in the first place?

> I ask @Maxez to check it on intel.

Do we have lighttpd on intel? For bigger loads we need to flood it with Apache Workbench or other similar tool (I know Janek used ab for stress-testing our network stack).

anglov:
I don't believe that as big load is needed to check if it's still an isse. I decrease stack size to 512B (as you told, before Janek change it was 16KB) 

If any error will occur on heavy load I suppose there is memory corruption somewhere else.

js6i:
> Hmm, maybe @js6i would be able to tell us why such a big stack was needed in the first place?

IIRC not actually needed, the problem was (is?) elsewhere.

nalajcie:
@js6i Thanks for the info! Not sure if the problem still persists, but we'll keep it in mind. For now we're trying to fit lwip into smaller target.

PR #10, opened 2020-09-01
Fix PPPoS driver and improve modem performance

Darchiv:
The timeout value for the initial `AT` command was unnecessarily multiplied. This made lwip trip over and block for several hours if the modem had not responded immediately. The retry count was also decreased because there is no point in waiting so long for a modem which has woken improperly.

Even though `AT+QCFG="autopdn",0` command is issued, the modem usually errors out when trying to establish a PPP connection (by error, `ERROR` return code for `CONNECT` command is meant). Addition of `AT+CGATT=0` resulted in 100% success rate of PPP connections, compared to ~30% without `AT+CGATT=0`.

### DIFF
--- a/drivers/pppos.c
+++ b/drivers/pppos.c
@@ -288,9 +288,9 @@ static int at_disconnect(int fd) {
 static int at_is_responding(int fd, int timeout_ms)
 {
 	int res;
-	int retry = 10;
+	int retry = 5;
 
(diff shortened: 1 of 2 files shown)


nalajcie (approved):
:+1:

anglov (approved):

PR #12, opened 2020-10-02
port: Remove pf handling if pf is not supported

anglov:


port/main.c
```diff
@@ -105,6 +105,7 @@ static int writeStatus(char *buffer, size_t bufSize, size_t offset)
 }
 
 
+#if HAVE_PF
```
nalajcie, line 108: I would suggest `#ifdef` or `#if defined(HAVE_PF)` as we're probably not compiling with `-Wundef`.

I would be happy to say that with `HAVE_*` macros, `#ifdef` is usually used, buuut - I've checked my `/usr/include` and - usually it's true (`#ifdef` for one `HAVE_*` flag, `#if defined(XXX) && defined(YYY)` for more complex ones), but there are some notable exceptions in well-known files (like `math.h`, or `stdlib.h`) - so I'm not forcing my point of view :).

Also: the support of HAVE_PF should be straightened up:
 - I see it's defined in `lwipopts.h` (I assume it should be done there, not in CFLAGS?), but this file is not explicitly included in the `port/main.c` (but included in other places checking for HAVE_PF)
 - other files use `#ifdef HAVE_PF` so either go back to `#ifdef` or switch to `#if everywhere`

port/main.c
```diff
@@ -167,11 +175,12 @@ static void mainLoop(void)
 			break;
 
 		case mtWrite:
+#if HAVE_PF
 				if (msg.i.io.oid.id == pf_oid.id)
 					msg.o.io.err = readPf(msg.i.data, msg.i.size, msg.i.io.offs);
 				else
+#endif
 					msg.o.io.err = -EINVAL;
```
nalajcie, line 183: maybe it could be set always before the ifdef? because the current code seems less-readable for me:
```c
    msg.o.io.err = -EINVAL;
#if defined(HAVE_PF)
    if (msg.i.io.oid.id == pf_oid.id)
	msg.o.io.err = readPf(msg.i.data, msg.i.size, msg.i.io.offs);
#endif
```
If that would be my code, I would request at least for braces to be added, eg:
```c
#if defined(HAVE_PF)
    if (msg.i.io.oid.id == pf_oid.id) {
	msg.o.io.err = readPf(msg.i.data, msg.i.size, msg.i.io.offs);
    } else
#endif
    {
	msg.o.io.err = -EINVAL;
    }
```

nalajcie (commented):
I'm assuming the motivation for this change is to lower the memory usage of the lwIP, correct?

port/main.c
anglov: As I told before if option is setted in lwipopts.h then it should be  #if something 
Agree that HAVE_PF will not be defaulted in lib_lwip BUT i think shouldn't be different handled than other option.

I enabled -Wundef after previous change :) 

IDK if this should be defined here instead of CFLAGS

anglov:
Change to #if elsewhere and fix style.

Also added next commit with fix no return warning.

port/main.c
nalajcie: Fine by me, sorry I forgot the `-Wundef` was already added.

I think we should maintain some convention, eg. options which are to be defined in `lwipopts.h` should be prefixed with `LWIP_` (or `LWIP_EXT_` if this our extension - like `LWIP_EXT_PF`?), all other `HAVE_*` defines are passed from Makefile (via `CFLAGS`).

Maybe it would be good to change it right now but, it's not a blocker against merging (as this topic is partially unrelated).

Please resolve this conversation if You don't want to change it.

port/main.c
```diff
@@ -233,4 +247,6 @@ int main(int argc, char **argv)
 		exit(1);
 
 	mainLoop();
+
+	return 0;
```
nalajcie, line 251: actually, if `mainLoop` finishes - it's probably an initialization error, so I would opt for `return 1` :).

port/main.c
anglov: TBH I think we have a big inconsistency – drivers defines by env and features changes by file lwipopts. I like LWIP_EXT_* so I changed it but I think we should define desired solution and make steps to do things in that way – maybe by that small changes also :)

nalajcie: Yeah, would be great to maintain some consistency. I hope that we will come to some conclusions when we will be using lwIP in many projects simultaneously (with different drivers and feature sets) - and come up with better solution then.

nalajcie (approved):

PR #13, opened 2020-10-12
pppos connect change

anglov:
New version hw/fw of modem changed their behavior – instead of hanging on AT_CONNECT_CMD now it returns error immediately if cannot do a connection yet. 

I added retrying connect command if doesn't timeouted. 

Also fixed some simple warnings

### DIFF
--- a/drivers/pppos.c
+++ b/drivers/pppos.c
@@ -114,12 +114,14 @@ static void serial_set_non_blocking(int fd)
 		log_error("%s() : fcntl(%d, O_NONBLOCK) = (%d -> %s)", __func__, fd, errno, strerror(errno));
 }
 
+#if 0
 static void serial_set_blocking(int fd)
 {
 	int flags = fcntl(fd, F_GETFL, 0);
 	if (fcntl(fd, F_SETFL, flags & ~O_NONBLOCK) < 0)
 		log_error("%s() : fcntl(%d, ~O_NONBLOCK) = (%d -> %s)", __func__, fd, errno, strerror(errno));
 }
+#endif
 
 #define WRITE_MAX_RETRIES 2
 static int serial_write(int fd, const u8_t* data, u32_t len)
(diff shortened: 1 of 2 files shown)


drivers/pppos.c
```diff
@@ -451,6 +452,7 @@ static void pppos_mainLoop(void* _state)
 {
 	pppos_priv_t* state = (pppos_priv_t*) _state;
 	int res;
+	int retry;
```
nalajcie, line 455: i would personally name it "reties" as this is the reties count and not binary "should I retry or not", but doesn't matter much

drivers/pppos.c
```diff
@@ -511,9 +513,13 @@ static void pppos_mainLoop(void* _state)
 		}
 #endif
 
-		if ((res = at_send_cmd(state->fd, AT_CONNECT_CMD, AT_CONNECT_CMD_TIMEOUT_MS)) != AT_RESULT_CONNECT) {
-			log_warn("failed to dial PPP, res=%d, retrying", res);
-			goto fail;
+		retry = AT_CONNECT_CMD_TIMEOUT_MS / PPPOS_CONNECT_CMD_RETRY_MS;
+		while ((res = at_send_cmd(state->fd, AT_CONNECT_CMD, AT_CONNECT_CMD_TIMEOUT_MS)) != AT_RESULT_CONNECT) {
+			if (retry-- <= 0 || res < 0) {
```
nalajcie, line 518: I would add at least comment that res will be < 0 in case timeout happened inside `at_send_cmd` - and/or **why** we would need to repeat the `at_send_cmd` at all (the generalized description of the pull request).

My initial thought was that the "old" modem was not handled properly (that we would retry the `at_send_cmd` X times even if each of the call would take `AT_CONNECT_CMD_TIMEOUT_MS`). I needed to check the implementation of `at_send_cmd` to understand the code - which should not be the case.

nalajcie (commented):
Seems ok, I've added some remarks regarding the readability which could be improved, but it's not a merge blocker for me.

anglov:
I made changes which you suggested.

nalajcie (approved):

PR #14, opened 2020-10-30
Handle SIM card with PIN

michalnowowiejski:


nalajcie:
**SHOULD NOT BE MERGED**
Incomplete solution to be deleted at a later time.

PR #16, opened 2021-01-12
Add armv7m7-imxrt106x LwIP target with device-to-device IoT communication driver

gerard5:
This changes brings an easy support for the IoT, device to device, connections on `armv7m7-imxrt106x` target. The `ppp over uart`/serial driver can be used with virtually any `phoenix-rtos` target that support uart, that being said, enabling easy porting of the `phoenix-rtos-lwip` to other IoT targets. **Developing IoT applications can be easier even on targets without ethernet, bluetooth and wifi**, but with simple point-to-point serial/uart null-modem connection.  Why `pppou` instead of already present `pppos` driver?  The answer is simple as `pppos` support modem GSM/GPRS/LTE communication (Hayes AT-command chat) I don't want to break that functionality, so `pppou` safetly derived for a null-modem implementation that doesn't depend on any other things than `phoenix-rtos-lwip` and  `phoenix-rtos` itself.

See it in action with debuging enabled:

https://user-images.githubusercontent.com/141153/104374934-e5f1a500-5522-11eb-95d8-438547535d2b.mp4

### DIFF
--- a/drivers/pppou.c
+++ b/drivers/pppou.c
@@ -1,0 +1,770 @@
+/*
+ * Phoenix-RTOS --- networking stack
+ *
+ * PPP over Serial driver
+ *
+ * Copyright 2018 Phoenix Systems
+ * Author: Marek Białowąs
+ *
+ * %LICENSE%
+ */
+
+#include "netif-driver.h"
+
+#include <netif/ppp/pppapi.h>
+
+#include <errno.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <stdarg.h>
+#include <string.h>
+#include <unistd.h>
+#include <strings.h>
+#include <sys/threads.h>
+#include <sys/msg.h>
+#include <syslog.h>
+
+#include <pppos_modem.h>
+
+enum {
+	CONN_STATE_DISCONNECTING,
+	CONN_STATE_DISCONNECTED,
+	CONN_STATE_CONNECTING,
+	CONN_STATE_CONNECTED,
+};
+
+typedef struct
+{
+	struct netif *netif;
+	ppp_pcb* ppp;
+
+	const char *serialdev_fn;
+	const char *serialat_fn;
+#if PPPOS_USE_CONFIG_FILE
+	const char *config_path;
+	char apn[64];
+#endif
+	int fd;
+
+	volatile int want_connected;
+	volatile int conn_state;
+	handle_t lock, cond;
+
+	uint32_t main_loop_stack[4096];
+} pppos_priv_t;
+
+#define COL_RED     "\033[1;31m"
+#define COL_CYAN    "\033[1;36m"
+#define COL_YELLOW  "\033[1;33m"
+#define COL_NORMAL  "\033[0m"
+
+#if 0
+#define log_debug(fmt, ...)     do { if (1) pppos_printf(state, fmt, ##__VA_ARGS__); } while (0)
+#define log_at(fmt, ...)     	do { if (1) pppos_printf(state, COL_CYAN fmt COL_NORMAL, ##__VA_ARGS__); } while (0)
+#define log_info(fmt, ...)      do { if (1) pppos_printf(state, COL_CYAN fmt COL_NORMAL, ##__VA_ARGS__); } while (0)
+#define log_warn(fmt, ...)      do { if (1) pppos_printf(state, COL_YELLOW fmt COL_NORMAL, ##__VA_ARGS__); } while (0)
+#define log_error(fmt, ...)     do { if (1) pppos_printf(state, COL_RED  fmt COL_NORMAL, ##__VA_ARGS__); } while (0)
+
+static void pppos_printf(pppos_priv_t *state, const char *format, ...)
+{
+	char buf[256];
+	va_list arg;
+
+	va_start(arg, format);
+	vsnprintf(buf, sizeof(buf), format, arg);
+	va_end(arg);
+
+	printf("lwip: ppp@%s %s\n", state->serialdev_fn, buf);
+}
+#else
+
+#define log_debug(fmt, ...) syslog(LOG_DEBUG, fmt, ##__VA_ARGS__)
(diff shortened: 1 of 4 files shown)


drivers/pppou.c
```diff
@@ -0,0 +1,627 @@
...
+
+		/* TODO: extend with other flags */
+	}
+
+	if (!state->serial_dev) {
+		log_error("no device");
+		return ERR_IF;
+	}
+
+	if (!state->serial_speed) {
```
kemonats, line 562:
```suggestion
	if (state->serial_speed < 0) {
```
Function `serial_speed_from_string(cfg))` returns `-1` if  it dosn't find string.

drivers/pppou.c
```diff
@@ -0,0 +1,627 @@
...
+#define PPPOU_READ_DATA_TIMEOUT_STEP_MS  10
+
+#define PPPOU_TRYOPEN_SERIALDEV_SEC      3
+#define PPPOU_CONNECT_RETRY_SEC          5
+#define PPPOU_CONNECT_CMD_RETRY_MS       500
+
+
+/****** serial handling ******/
+
+speed_t serial_speed_from_string(const char str[])
```
kemonats, line 99:
```suggestion
speed_t serial_speed_from_string(const char *str)
```
I think it will be better.

drivers/pppou.c
gerard5: I don't agree. See line 543 where it's actually checked  `speed >= 0` and if true then `state->serial_speed = speed` is set, so until `serial_speed_from_string()` doesn't find proper speed `state->serial_speed` is set to zero (with memset), I assume zero as incorrect value because of `B0` value (`termios.h`).
The zero baud rate, `B0`, is used to terminate the connection.

drivers/pppou.c
gerard5: This way `char str[]` is denoted as C style string parameter _(array of characters `\0` terminated)_ , what You suggest is a pointer to the very first signed character of a memory - nearly the same, but not exactly. See also, I'm consistent with `serial_open()`, where `devname` is also a string there. String names are decayed to pointers, hence, You can use pointers to manipulate elements of the string. The difference is when it comes to initializers, then of course assigning `char str[]` a pointer is invalid.
As absolute classic:  `int main(int argc, char *argv[], char *envp[])` of course one can write this as `main(int argc, char **argv, char **envp)`, nearly the same, but ... I'm tied to the classic.

drivers/pppou.c
kemonats: Ok. You're right.

kemonats:
The declaration of the `strcm()` function is as follows:
```
extern int strcmp(const char *str1, const char *str2);
```
This is your code and you decide.

drivers/pppou.c
```diff
@@ -23,109 +22,199 @@
...
+	for (cnt = 0; (ret = lookup(devname, NULL, &oid)) < 0; cnt++) {
+		usleep(200 * 1000);
+		if (cnt > 3)
+			return ret;
+	}
+
+	if ((fd = open(devname, O_RDWR | O_NOCTTY | O_NONBLOCK)) < 0)
+		return fd;
+
+	if ((ret = serial_set_blocking(fd, 0)) < 0)
```
nalajcie, line 171: Why setting `O_NONBLOCK` twice? (it was already set with `open` flags)?

drivers/pppou.c
```diff
@@ -619,152 +479,167 @@ static int pppos_netifUp(pppos_priv_t *state)
...
+		ppp_set_netif_statuscallback(state->ppp, pppou_statusCallback);
 	}
 
-	beginthread(pppos_mainLoop, 4, (void *)state->main_loop_stack, sizeof(state->main_loop_stack), state);
-
-	return ERR_OK;
-}
+	beginthread(pppou_mainLoop, 4, (void *)state->main_loop_stack, sizeof(state->main_loop_stack), state);
 
+	for (retries = 3; retries > 0; retries--) {
```
nalajcie, line 607: I assume this is temporary because of lack of real `ifconfig` on the target platform (usually I would not like for netif initialization to be unnecessarily blocking)

drivers/pppou.c
```diff
@@ -23,109 +22,199 @@
...
+		{ "1800", B1800 },     { "19200", B19200 },   { "200", B200 },
+		{ "230400", B230400 }, { "2400", B2400 },     { "300", B300 },
+		{ "38400", B38400 },   { "460800", B460800 }, { "4800", B4800 },
+		{ "50", B50 },         { "57600", B57600 },   { "600", B600 },
+		{ "75", B75 },         { "9600", B9600 },
+	};
+
+	struct serial_speed_s *sptr;
+
+	if (str && (sptr = bsearch(&str, stab, sizeof(stab) / sizeof(stab[0]), sizeof(stab[0]), _speedcmp)))
```
nalajcie, line 126: a little bit of over-engineering, simple `for` loop would be probably enough (or even a switch-case block) :sunglasses:

nalajcie (approved):
Looks good, thanks for the description in the `pppou.c` header regarding how to set up the connection.

Left some notes, none of them are a merge blocker.

I'm wondering if we should start writing some docs in the [phoenix-rtos-doc](https://github.com/phoenix-rtos/phoenix-rtos-doc/tree/master/lwip) repo when we're adding new functionality. If You have the time, feel free to add short description with sample usage there (probably the same text You've put in `pppou.c` header comment but in markdown).

drivers/pppou.c
gerard5: That's right.

drivers/pppou.c
gerard5: Ack.

drivers/pppou.c
gerard5: That's right, because serial_set_blocking is a static function and only here it is used I will remove a call and the definition. push-force the patch.

nalajcie:
Merging this now, we might add documentation in `phoenix-rtos-doc` in the future.

PR #17, opened 2021-01-29
port: Fix ioctl warnings

MaciejPurski:
This PR, combined with https://github.com/phoenix-rtos/libphoenix/pull/76 fixes loads of ugly warnings when building lwip, e.g.:
Screenshot from 2021-01-29 12-19-47

I know, the solution is hacky, but it seems like the simplest way to fix it without changing the whole idea of using both `libphoenix` and `lwip` headers in `phoenix-rtos-lwip/port/sockets.c`. However, I'm opened to comments and other ideas.

### DIFF
--- a/port/sockets.c
+++ b/port/sockets.c
@@ -9,6 +9,8 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+#include <sys/ioctl.h>
+
 #define ifreq lwip_ifreq
 #include <lwip/netdb.h>
 #include <lwip/sockets.h>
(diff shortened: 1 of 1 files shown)


port/sockets.c
```diff
@@ -18,6 +20,7 @@
 #include <lwip/dhcp.h>
 #include <lwip/prot/dhcp.h>
 #undef ifreq
+#undef IFNAMSIZ
```
nalajcie, line 23: this shouldn't probably be needed, but it depends on "untangling libphoenix includes" task, so I'm ok with merging this as-is.

nalajcie (approved):

PR #20, opened 2021-03-24
pppou: Fix, set default route via ppp interface

gerard5:
With this fix, the `pppou` driver by default will add the `default-route`. When `default-route` is not to be added  `nodefault` configuration flag should be used

Add default route e.g.
```
app flash0 -x @lwip;pppou:/dev/uart3:115200:up xip1
```

Do not add default route e.g.
```
app flash0 -x @lwip;pppou:/dev/uart3:115200:up:nodefault xip1
```


MaciejPurski:
hw tested on imxrt1064, besmart board

PR #22, opened 2021-04-29
drivers/pppou: Enable using peer's DNS address

MaciejPurski:
This PR enables fetch DNS address from a peer, when connecting via PPPoU.
An exemplary Linux command, which sets the DNS address for the peer:
`pppd /dev/ttyUSB0 460800 10.0.0.1:10.0.0 ms-dns 8.8.8.8 lock local nodetach noauth debug dump nocrtscts maxfail 0 holdoff 1`

### DIFF
--- a/drivers/pppou.c
+++ b/drivers/pppou.c
@@ -11,6 +11,8 @@
 
 #include "netif-driver.h"
 
+#include <lwip/dns.h>
+
 #include <netif/ppp/pppapi.h>
 
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


drivers/pppou.c
```diff
@@ -585,7 +592,9 @@ static int pppou_netifInit(struct netif *netif, char *cfg)
 		}
 		netif->flags &= ~NETIF_FLAG_UP;
 		ppp_set_netif_statuscallback(state->ppp, pppou_statusCallback);
-
+#if LWIP_DNS
+		ppp_set_usepeerdns(state->ppp, 1);
```
gerard5, line 596: You can add similar logic to the one used with the `ppp_set_default ()` route below, e.g. by using the `CFG_FLAG_NO_DNS` flag (configuration option`"nodns"`), this way DNS will be enabled by default but can be easily disabled in pppou configuration. Of course, between ifdefs `LWIP_DNS`.

nalajcie (approved):

PR #32, opened 2021-07-13
Ziemleszcz/ipsec

ziemleszcz:
This commit prepares IPSEC build, adds some extra info to /dev/ifstatus and creates new file /dev/ipsecdev to monitor ethernet cable status.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

Requires (to work, not to build):
https://gerrit.phoesys.com/c/dataro/+/2864/1
https://gerrit.phoesys.com/c/dataro/+/2865/1

- [x] This PR needs additional PRs to work (list the PRs, preferably in merge-order).

### DIFF
--- a/port/main.c
+++ b/port/main.c
@@ -18,8 +18,9 @@
 #include <sys/msg.h>
 #include <posix/utils.h>
 
-#include<lwip/sockets.h>
-#include<lwip/inet.h>
+#include <lwip/sockets.h>
+#include <lwip/inet.h>
+#include <lwip/dns.h>
 
 #include <stdio.h>
 #include <stdlib.h>
@@ -47,6 +48,15 @@
 		size -= write; \
 		buffer += write; \
 	} while (0)
+
+
+#ifdef LWIP_LINKMONITOR_DEV
+static struct {
+	int disconnected;
+	int reconnected;
+	int last_link;
+} main_common;
+#endif
 
 
 static int writeRoutes(char *buffer, size_t bufSize, size_t offset)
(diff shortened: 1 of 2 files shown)


port/main.c
```diff
@@ -121,10 +133,69 @@ static int writeStatus(char *buffer, size_t bufSize, size_t offset)
 			SNPRINTF_APPEND("%2s%d_media=%s\n", netif->name, netif->num, drv->media(netif));
 	}
 
+#if LWIP_DNS
+	for (i = 0; i < LWIP_DHCP_MAX_DNS_SERVERS; ++i)
+		SNPRINTF_APPEND("dns_%u=%s\n", i, inet_ntoa(*dns_getserver(i)));
```
nalajcie, line 138: not connected to this PR: what will happen if we would setup 2 ethernet ports to DHCP? I think everything will work apart from (potentially) DNS servers?

port/main.c
```diff
@@ -49,6 +50,13 @@
 	} while (0)
 
 
+#if LWIP_IPSEC
+static int ipsecdev_disconnected;
```
nalajcie, line 54: I believe our coding convention is to put all global variables to one "common" struct.

I like the simplicity of this solution, but the naming is misleading (link monitoring is not connected to IPSec in any way, just the interface name is taken from `LWIP_IPSEC_DEV`).

I think we could either:
 - provide separate define for monitoring device (so the functionality would not be under `LWIP_IPSEC`)
 - add support for writing device name in runtime (so to start monitoring the device I would do `echo -n "en1" > /dev/linkmonitor`) and than start reading the file?

nalajcie (changes requested):
I think we should change the naming (drop link monitoring and IPSec dependency). Other than that - I like the simplicity of the approach, regardless of the need to poll for the changes.

We could think of adding blocking read here (storing readers on wait queue and releasing them on link status change) in the future - but not right now as blocking reads can't be interrupted/canceled on our OS yet.

port/main.c
ziemleszcz: DNS servers are overwritten by DHCP .

nalajcie: thx for the info

nalajcie (approved):
LGTM.

PR #34, opened 2021-07-19
dev: lwip /dev/* entries refactor

ziemleszcz:
Legacy code was incorrectly handling EOF (last read returns ERANGE)
and could potentially return corrupted data (state change between
reads).

JIRA: DTR-91

## Motivation and Context
Required for route command. This version fixes EOF, can handle all buffer sizes and will not return corrupted data. It's not a perfect solution because access is limited to 1 reader/writer.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/port/main.c
+++ b/port/main.c
@@ -16,6 +16,7 @@
 #include <lwipopts.h>
 
 #include <sys/msg.h>
+#include <sys/minmax.h>
 #include <posix/utils.h>
 
 #include <lwip/sockets.h>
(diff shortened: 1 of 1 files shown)


port/main.c
```diff
@@ -33,38 +34,68 @@
...
+	size_t size;
+	int n;
+
+	if (flags & (O_WRONLY | O_RDWR))
+		return -EACCES;
 
+	if (main_common.route.busy)
+		return -EBUSY;
+
+	mutexLock(rt_table.lock);
```
nalajcie, line 95: please remove fixme in line 102

port/main.c
```diff
@@ -33,38 +34,68 @@
...
 
-#ifdef LWIP_LINKMONITOR_DEV
 static struct {
-	int disconnected;
-	int reconnected;
-	int last_link;
-} main_common;
+	struct {
+		int busy;
+		char buf[1024];
```
nalajcie, line 55: This is kindof wasteful on small tagets (`stm32l1`). Maybe we can avoid it somehow (either by ifdef or some other mechanism?)
@anglov, do You use these files at all? Maybe we can put them under single ifdef? (`LWIP_DEV_FILES` or something like that?)

port/main.c
```diff
@@ -102,16 +133,61 @@ static int writeRoutes(char *buffer, size_t bufSize, size_t offset)
 			0, 0); /* Window, IRTT */
 	}
 
-	return bufSize - size;
+	main_common.route.busy = 1;
+	main_common.route.len = buf - main_common.route.buf;
+
+	mutexUnlock(rt_table.lock);
```
nalajcie, line 139: technically could be done earlier but I know it's because of `goto` :p

nalajcie (commented):
The code is better structured now, I like that You've removed the potential concurrency issues. Maybe it would be worth to move the dev file handling part to separate file in the future as it seems to grow in size :).

My concerns:
 - it will increase memory usage due to static buffers (might be a dealbreaker for `stm32l1` - not giving approval until @anglov would say it's ok)
 - we try to avoid `goto` (because of our coding convention), I don't see a clear solution here (apart from removing the macro and writing the code inline - which can be done as it's much shorter now). To be considered eventually.

Apart from that, LGTM. Haven't tested on real targets.

port/main.c
anglov: Thanks for metioning me here:
- I think it's ok for now to slighthly increase memory usage (it shouldn't break anything for now)
- Whole /dev files increase binary size and memory usage. So I would like to put those under ifdefs (or single ifdef as you suggest) as it not needed for small targets that don't have to have routing information, packet filter and even netstatus via /dev interface
- I'm not sure if putting it under one define is ok. Some configuration and status driver will be needed for PPP. It could be done by separate /dev interface. It have to be discussed how to achieve this.

nalajcie (approved):

PR #63, opened 2022-04-22
port/threads: sys_thread_join fix

ziemleszcz:
JIRA: DTR-290

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/port/threads.c
+++ b/port/threads.c
@@ -33,7 +33,6 @@ static struct {
 	char collector_stack[512] __attribute__((aligned(8)));
 	rbtree_t threads;
 	handle_t lock;
-	handle_t join_lock;
 	handle_t join_cond;
 } global;
 
@@ -96,9 +95,7 @@ static void thread_waittid_thr(void *arg)
 			free(data->stack);
 			free(data);
 
-			mutexLock(global.join_lock);
-			condSignal(global.join_cond);
-			mutexUnlock(global.join_lock);
+			condBroadcast(global.join_cond);
 		}
 		else {
 			mutexUnlock(global.lock);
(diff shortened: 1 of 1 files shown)


port/threads.c
```diff
@@ -96,9 +95,9 @@ static void thread_waittid_thr(void *arg)
 			free(data->stack);
 			free(data);
 
-			mutexLock(global.join_lock);
-			condSignal(global.join_cond);
-			mutexUnlock(global.join_lock);
+			mutexLock(global.lock);
```
nalajcie, line 98: Is it sensible to free the lock in line 92 just to reacquire it here?

Also: `mutex_unlock` could probably be moved outside of the if-else?

port/threads.c
```diff
@@ -96,9 +95,9 @@ static void thread_waittid_thr(void *arg)
...
+			condBroadcast(global.join_cond);
+			mutexUnlock(global.lock);
```
agkaminski, line 100: condSignal, condBroadcast does CPU reschedule, it's way better to call these without mutex.

condBroadcast -> reschedule -> mutexLock -> reschedule -> mutexUnlock -> reschedule
vs
condBroadcast -> reschedule -> mutexLock

agkaminski: Both signal and broadcast execute this function: https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/8ee6add2b00884ea372506e542c74c53efc6a4f1/proc/threads.c#L1211
which does the reschedule if there was a thread on the queue

agkaminski (approved):

PR #78, opened 2022-09-29
libsdio: use libsdio in WHD HAL

Arthur-Sebastian:
JIRA: RTOS-212

## Types of changes

## How Has This Been Tested?

- [ ] Tested by hand on: (list targets here).

## Checklist:

## Special treatment

### DIFF
--- a/wi-fi/hal/cyhal_sdio.c
+++ b/wi-fi/hal/cyhal_sdio.c
@@ -4,766 +4,128 @@
  * LwIP Wi-Fi
  *
  * Copyright 2021 Phoenix Systems
- * Author: Ziemowit Leszczynski
+ * Author: Ziemowit Leszczynski, Artur Miller
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
 #include "cyhal_sdio.h"
-#include "cyhal_utils.h"
-#include "cyabs_rtos.h"
-#include "cy_log.h"
-#include "physmmap.h"
-
-#include <string.h>
+#include "cyhal_gpio.h"
+#include <stdio.h>
 #include <unistd.h>
-#include <sys/mman.h>
-#include <sys/interrupt.h>
 
-
-#define USDHC2_ADDR 0x2194000
-#define USDHC2_IRQ  (32 + 23)
-
-#define DMA_BUFFER_SIZE 2048  // should be large enough to hold whole Wi-Fi frame
-
-enum { ds_addr = 0,
-	blk_att,
-	cmd_arg,
-	cmd_xfer_typ,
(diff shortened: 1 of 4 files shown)


wi-fi/hal/cyhal_sdio.c
```diff
@@ -4,766 +4,128 @@
...
-	uint32_t val, cmd, mix, blk, block_mode, count;
-
-	block_mode = !!(argument & (1 << 27));
-	count = argument & 0x1ff;
-
-	if (response)
+	/* WHD only uses send cmd api call for
+	 * transfers after init commands which are
+	 * issued by sdio_init function, hence
+	 * just indicate success and exit */
```
MaciejPurski, line 84: I'm afraid we can't do it this way - assuming a specifiic behaviour of the upper layer, in this case the whd driver. It might change in the future, we need to keep it generic still. Unfortunately I see no other option here than exporting `sdio_sendCmd`, which is now static, and maybe check if the cmd is CYHAL_SDIO_CMD_IO_RW_DIRECT and use transferDirect function in this case and sendCmd otherwise.

wi-fi/hal/cyhal_sdio.c
```diff
@@ -4,766 +4,128 @@
...
-		if (res == CY_RSLT_SUCCESS)
-			break;
-
-		cy_log_msg(CYLF_SDIO, CY_LOG_ERR, "repeating SDIO bulk transfer cmd (cnt=%d)\n", i);
-
-		usleep(1000);
-	}
-
+	size_t len;
+	sdio_cmd_arg_t arg;
```
MaciejPurski, line 103: It could be done in an even smarter way:
```suggestion
	whd_bus_sdio_cmd53_argument_t arg = (whd_bus_sdio_cmd53_argument_t)argument;

```

wi-fi/hal/cyhal_sdio.c
```diff
@@ -4,766 +4,128 @@
...
 
-		while (1) {
-			finish = sdio_common.irq_thread_finish;
-			if (finish)
-				break;
 
-			val = *(sdio_common.base + int_status);
-			if (val & (1 << 8))
-				break;
+cyhal_sdio_irq_handler_t conv;
```
MaciejPurski, line 42: At least it should be static. But I think the most portable solution would be to add a structure for keeping the sdio context with only this value here and make use of the (cyhal_sdio_t *obj) pointer, which gets passed to basically every function.

MaciejPurski (commented):
Some changes need to be done and we need a further discussion on some of the libsdio API changes.

PR #89, opened 2024-02-01
drivers/ephy: more dynamic support for KSZ8081 RNA/RND variants

nalajcie:
## Description

- drivers/ephy: add MACPHY self-test
  Add MACPHY test using local loopback mode

- drivers/ephy: add support for alternative configurations
  - read / pass board revision to ephy_init
  - add option for custom init hook to select alternative phy configuration
  - rewrite RND/RNA ext config logic override

## Motivation and Context
 JIRA: DTR-480

Needed for DTRv2

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: DTR, DTRv2

## Checklist:

## Special treatment

### DIFF
--- a/drivers/Makefile
+++ b/drivers/Makefile
@@ -1,7 +1,9 @@
 # Makefile for lwip drivers to be included
 
-# TODO: make every driver a separate component?
-ifneq ($(EPHY_KSZ8081RND),)
+ifeq ($(EPHY_KSZ8081),RNA)
+  LOCAL_CFLAGS := -DEPHY_KSZ8081RNA
+endif
+ifeq ($(EPHY_KSZ8081),RND)
   LOCAL_CFLAGS := -DEPHY_KSZ8081RND
 endif
 
--- a/drivers/ephy.c
+++ b/drivers/ephy.c
@@ -280,8 +280,49 @@ int ephy_enableLoopback(eth_phy_state_t *phy, bool enable)
 	return 0;
 }
 
-int ephy_init(eth_phy_state_t *phy, char *conf, link_state_cb_t cb, void *cb_arg)
+
+/* Try to set alternative MAC PHY config (alternative configurations within the same PHY ID).
+ * returns:
+ *   > 0 if alternative config has been set
+ *     0 if no alternative config with this ID
+ *   < 0 if alternative config setting has failed
+ */
+static int ephy_setAltConfig(eth_phy_state_t *phy, int cfg_id)
 {
+	/* NOTE: assuming KSZ8081 RNA/RND PHY! */
+
+	/* CFG id:
+	 * 0: KSZ8081 RND with 50 MHz RMII input clock (PHY_CTRL2[7] = 0)
+	 * 1: KSZ8081 RNA with 50 MHz RMII input clock (PHY_CTRL2[7] = 1)
+	 */
+
+
+	/* try to set alternative MII clock frequency */
+	uint16_t phy_ctrl2 = ephy_reg_read(phy, 0x1f);
+
+	if (cfg_id == 0) {
+		phy_ctrl2 &= ~(1 << 7);
+	}
+	else if (cfg_id == 1) {
+		phy_ctrl2 |= (1 << 7);
+	}
+	else {
+		return 0; /* unknown config ID */
+	}
+
+	ephy_reg_write(phy, 0x1f, phy_ctrl2);
+	if (ephy_reg_read(phy, 0x1f) != phy_ctrl2) {
+		printf("ephy: failed to set alternative clock\n");
+		return -1;
+	}
+
+	return 1;
+}
+
+
+int ephy_init(eth_phy_state_t *phy, char *conf, uint8_t board_rev, link_state_cb_t cb, void *cb_arg)
+{
+	(void)board_rev;
 	uint32_t phyid;
 	int err;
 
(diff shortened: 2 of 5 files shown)


drivers/Makefile
```diff
@@ -1,7 +1,9 @@
 # Makefile for lwip drivers to be included
 
-# TODO: make every driver a separate component?
-ifneq ($(EPHY_KSZ8081RND),)
+ifeq ($(EPHY_KSZ8081),RNA)
+  LOCAL_CFLAGS := -DEPHY_KSZ8081RNA
+endif
+ifeq ($(EPHY_KSZ8081),RND)
```
agkaminski, line 6: Perhaps it would be better to remove RND/RNA compile time distinction, as we now support dynamic detection anyway?

nalajcie: We support dynamic **selection** by custom hook. Dynamic detection proved to be unreliable in some circumstances (software reboot, cable in the port, etc.) so finally it's not implemented (maybe after some more experimentation it could be done more consistently). The `do {} while (0)` is a remnant of `do {} while (ephy_setAltConfig(cfg++) > 0)`.

We sometimes still would be able to force one of the possibilites at a compile time (eg. during production when board revision is not yet burned into fuses). That's why I intentionally left it as an env variable (not a `board_config.h` const) - to be able to easily force it from command line when building.

drivers/imx6-enet.c
```diff
@@ -793,6 +819,134 @@ static void enet_setLinkState(void *arg, int state)
...
+		memset(p->payload, 0, ETH_PAD_SIZE);
+		pbuf_take_at(p, TEST_PACKET, TEST_PACKET_LEN, ETH_PAD_SIZE);
+
+		/* try to send and receive packets */
+		mutexLock(priv->selfTest.rx_lock);
+		priv->selfTest.rx_cnt = 0;
+		if (enet_netifOutput(netif, p) != ERR_OK) { /* frees pbuf internally */
+			enet_printf(priv, "failed to send test packet");
+			ret = -1;
+			break;
```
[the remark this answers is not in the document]
nalajcie: thx, done.

Never actually happened while testing, so I've missed this during final refactor :).

drivers/ephy.c
```diff
@@ -280,9 +356,16 @@ int ephy_init(eth_phy_state_t *phy, char *conf, link_state_cb_t cb, void *cb_arg
 	/* make address 0 not broadcast, disable NAND-tree mode */
 	ephy_reg_write(phy, 0x16, 0x0202);
 
-#ifndef EPHY_KSZ8081RND
-	/* 50MHz RMII clock; keep auto-MDI-X */
-	ephy_reg_write(phy, 0x1f, 0x8180);
+	ephy_setAltConfig(phy, 0); /* KSZ8081 RND - default config */
```
ziemleszcz, line 359: Move to `#else` section after `#elif defined(EPHY_KSZ8081RND)`?

agkaminski (dismissed):
LGTM

drivers/imx6-enet.c
```diff
@@ -838,9 +993,16 @@ static int enet_netifInit(struct netif *netif, char *cfg)
...
 
 	if (cfg) {
-		err = ephy_init(&priv->phy, cfg, enet_setLinkState, (void*) priv->netif);
+		uint8_t board_rev = enet_readBoardRev();
+
+		err = ephy_init(&priv->phy, cfg, board_rev, enet_setLinkState, (void *)priv->netif);
 		if (err)
 			enet_printf(priv, "WARN: PHY init failed: %s (%d)", strerror(err), err);
+
+		err = enet_phySelfTest(netif);
```
ziemleszcz, line 1002: Does it make sense to run selftest if `ephy_init` fails?

drivers/imx6-enet.c
```diff
@@ -793,6 +819,135 @@ static void enet_setLinkState(void *arg, int state)
...
+	enet_priv_t *priv = netif->state;
+	int err;
+
+	if ((err = create_mutexcond_bulk(SELFTEST_RESOURCES(priv))) != 0) {
+		return err;
+	}
+
+	/* setup self-test (local loopback mode & force linkup) */
+	if (ephy_enableLoopback(&priv->phy, true) < 0) {
+		return -1;
```
ziemleszcz, line 883: Resources are not destroyed.

drivers/imx6-enet.c
```diff
@@ -793,6 +819,135 @@ static void enet_setLinkState(void *arg, int state)
...
+		is_valid_pkt = false;
+	}
+	pbuf_free(p);
+
+	mutexLock(priv->selfTest.rx_lock);
+	if (is_valid_pkt) {
+		priv->selfTest.rx_cnt += 1;
+	}
+	mutexUnlock(priv->selfTest.rx_lock);
+	condBroadcast(priv->selfTest.rx_cond);
```
ziemleszcz, line 865: Call before unlocking mutex?

nalajcie: Yeah, the lock/cond pattern should probably be done as you've described but it turns out that it's slower on our OS (cond broadcast would result in immediate reschedule).

Further investigation proved it's not easy (especially in SMP) to optimize this code path and it usually isn't in various pthread implementations.

ziemleszcz: OK, in that case you don't need to lock the mutex if `is_valid_pkt` is false.

nalajcie: renamed `rx_cnt` to `rx_valid` and explicitly mark receiving invalid packet to break `condWait` earlier
(testing showed it's common to receive broken packet when PHY configuration is invalid)

drivers/imx6-enet.c
nalajcie: thx, done. Added explicit loopback mode disable just for safety.

drivers/imx6-enet.c
nalajcie: Probably not but I wanted to avoid increasing cyclomatic complexity. Added explicit early return.

drivers/imx6-enet.c
```diff
@@ -223,6 +232,23 @@ static inline uint8_t get_byte(uint32_t v, int i)
...
+
+	if (va == MAP_FAILED)
+		return 0;
+
+	/* note: keep in sync with imx6ull-otp */
+	res = va[0x100 + 4 * 0x27];
+
+	physunmap(va, 0x1000);
+
+	return (res >> 24);
```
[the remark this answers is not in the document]
nalajcie: done.

ziemleszcz (approved):
LGTM

agkaminski (approved):

PR #96, opened 2024-08-26
drivers/imx-enet: Add support for ethernet in imxrt1064-evk

julianuziemblo:
## Description
Add Ethernet support for iMX RT106x, merging the code with existing driver for iMX 6ULL in `drivers/imx-enet.c`.

## Motivation and Context

[https://jira.phoenix-rtos.com/browse/RTOS-507](https://jira.phoenix-rtos.com/browse/RTOS-507)

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `armv7a7-imx6ull-evk`, `armv7m7-imxrt106x-evk`, `armv7a7-imx6ull-dataro`, `ia32-generc-qemu`.

## Checklist:

- [x] My change requires a change to the documentation.

## Special treatment

    - https://github.com/phoenix-rtos/phoenix-rtos-project/pull/1134
    - https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/601
    - https://github.com/phoenix-rtos/plo/pull/355


drivers/imx-enet.c
```diff
@@ -820,78 +1180,92 @@ static void enet_setLinkState(void *arg, int state)
...
+#define _TP_SRC     "ssssss"
 #define _TP_ETHTYPE "\x05\xDD" /* eth frame type 0x05DD is undefined */
 #define _TP_10DIG   "0123456789"
-#define TEST_PACKET "ddddddssssss" _TP_ETHTYPE \
+#define TEST_PACKET _TP_DST _TP_SRC _TP_ETHTYPE \
 	_TP_10DIG _TP_10DIG _TP_10DIG _TP_10DIG _TP_10DIG _TP_10DIG _TP_10DIG
-#define TEST_PACKET_LEN (sizeof((TEST_PACKET)) - 1)
-
+#define TEST_PACKET_LEN             (sizeof((TEST_PACKET)) - 1)
+#define ENET_PHYSELFTEST_RX_TIMEOUT (100 * 1000) /* TX -> RX takes ~4ms, wait for 100ms just to be sure */
```
nalajcie, line 1191: comment in the only place it's being used has better readability, please revert the change

drivers/imx-enet.c
```diff
@@ -723,93 +1039,137 @@ static int enet_pinConfig(enet_priv_t *state)
...
-			priv->mmio->RCR &= ~ENET_RCR_RMII_10T;
+		speed = ephy_linkSpeed(&priv->phy, NULL);
+
+		switch (speed) {
+			case 10:
+				priv->mmio->RCR |= ENET_RCR_RMII_10T;
+				break;
+			case 100:
+				priv->mmio->RCR &= ~ENET_RCR_RMII_10T;
+				break;
```
nalajcie, line 1172: MISRA: every `switch` needs to have `default`

drivers/imx-enet.c
```diff
@@ -820,78 +1180,92 @@ static void enet_setLinkState(void *arg, int state)
...
+	enet_state_t *state = netif->state;
 
 	bool is_valid_pkt = true;
 
 	/* verify contents */
 	if ((p->len != TEST_PACKET_LEN + ETH_PAD_SIZE)) {
-#ifdef ENET_VERBOSE
-		enet_printf(priv, "self-test RX: invalid packet length");
-#endif
+		enet_printf(state, "self-test RX: invalid packet length");
```
nalajcie, line 1205: not sure if shouldn't still be under debug/verbose

drivers/imx-enet.c
```diff
@@ -465,121 +592,165 @@ static void enet_irq_thread(void *arg)
...
 }
 
 
-static uint16_t enet_mdioIO(enet_priv_t *state, unsigned addr, unsigned reg, unsigned val, int read)
+static uint16_t enet_mdioIO(enet_state_t *state, unsigned addr, unsigned reg, uint16_t val, unsigned op)
 {
-	state->mmio->EIR = ENET_IRQ_MII;
-	if (!MDC_ALWAYS_ON)
-		state->mmio->MSCR = state->mscr;
+#if MDIO_DEBUG
```
nalajcie, line 689: not sure about `MDC_ALWAYS_ON` functionality - but if was previously enabled and now is missing

drivers/ephy.c
```diff
@@ -23,151 +24,270 @@
...
+	}
+#if defined(__CPU_IMX6ULL)
+	else
+#endif
+	{
+#if EPHY_DEBUG
+		ephy_printf(phy, "ephy_reset: start software reset...");
+#endif
+		/* software reset procedure */
+		ephy_reg_set(phy, EPHY_00_BASIC_CONTROL, (1 << 15));
```
nalajcie, line 116: Uhh, shouldn't we reset all other fields to 0 when asserting reset?

drivers/ephy.c
```diff
@@ -23,151 +24,270 @@
...
+}
+
+
+static inline void ephy_reg_set(eth_phy_state_t *phy, uint16_t reg, uint16_t val)
+{
+	ephy_reg_write(phy, reg, ephy_reg_read(phy, reg) | val);
+}
+
+
+static inline void ephy_reg_clr(eth_phy_state_t *phy, uint16_t reg, uint16_t val)
```
nalajcie, line 70: never used, remove

drivers/ephy.c
```diff
@@ -23,151 +24,270 @@
...
 	mdio_write(phy->bus, phy->addr, reg, val);
-	ephy_reg_read(phy, reg);
+	uint16_t ret __attribute__((unused)) = ephy_reg_read(phy, reg);
+#if MDIO_DEBUG
+	ephy_printf(phy, "write: val from reg: 0x%04x", ret);
+#endif
+}
+
+
+static inline void ephy_reg_set(eth_phy_state_t *phy, uint16_t reg, uint16_t val)
```
nalajcie, line 64: used only twice, not sure if needed in any of these cases

nalajcie (changes requested):
Partial review as current commit structure prevents efficient code review (the browser crashed with the amount of changes/comments in single commit :)).

Short summary:
 - fix commit message titles (see `git history` for examples), at least prefix them with module name (also include in commit message bodies what / why was changed), add missing JIRA ID
 - clearly split the changes into separate atomic commits (eg. changing type name is not bringing new functionality and should be a separate change)
 - fix coding style (according to our coding convention / MISRA as described in some comments)
 - please test the code after changes on imx6ull-evk and DTR project (from CI I see that something is broken there), check also if the performance is not degraded

nalajcie:
converting to draft until it will be ready for next review

drivers/imx-enet.c
```diff
@@ -32,247 +35,324 @@
...
+typedef enet_legacy_desc_t enet_buf_desc_t;
 #endif
 
 typedef struct
 {
 	volatile struct enet_regs *mmio;
 
 	struct netif *netif;
-	unsigned drv_exit;
+	_Atomic volatile unsigned drv_exit;
```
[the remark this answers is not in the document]
julianuziemblo: It is used as atomic e.g. in function `enet_irq_handler`, so it should be marked as such

drivers/imx-enet.c
```diff
@@ -311,17 +396,26 @@ static size_t enet_nextRxBufferSize(const net_bufdesc_ring_t *ring, size_t i)
...
 #endif
 
-	if (desc->flags & ENET_DESC_OWN)
+	if (atomic_load(&desc->flags) & ENET_DESC_FLAG_READY)
 		return 0;
 
 	sz = desc->len;
-	if (!sz)	// FIXME: hw bug?
+	if (!sz) {
+		printf("WARNING: This message indicates a potential HW bug:\n");
```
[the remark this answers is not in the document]
julianuziemblo: Cannot call `enet_printf`, because this function doesn't have the access to the `enet_state_t` instance

drivers/imx-enet-regs.h
```diff
@@ -305,7 +305,7 @@ struct enet_regs {
 typedef struct
 {
 	uint16_t len; /* for last frag = whole frame size? (TBV) */
-	uint16_t flags;
+	_Atomic volatile uint16_t flags;
```
nalajcie, line 308: - this is read/written by the hardware, so changing this to atomic doesn't do the expected thing
- `volatile` is also not needed, as we're always declare the full struct as `volatile`

drivers/imx-enet-regs.h
```diff
@@ -314,7 +314,8 @@ typedef struct
 	/* first 3 same as enet_legacy_desc_t */
 	union {
 		struct {
-			uint16_t len, flags;
+			uint16_t len;
+			_Atomic volatile uint16_t flags;
```
nalajcie, line 318: - this is read/written by the hardware, so changing this to atomic doesn't do the expected thing
- `volatile` is also not needed, as we're always declare the full struct as `volatile`

drivers/imx-enet.c
```diff
@@ -513,9 +513,11 @@ static void enet_irq_thread(void *arg)
 {
 	enet_state_t *state = arg;
 	size_t rx_done = 0;
+	unsigned exit;
 
 	mutexLock(state->irq_lock);
-	while (state->drv_exit == 0) {
+	exit = atomic_load(&state->drv_exit);
+	while ((exit & EV_BUS_ERROR) == 0) {
```
nalajcie, line 520: why checking only one bit? 
I don't care that currently only this bit is being set, looking at this code I would assume there are other non-fatal errors (incorrectly passed by `drv_exit` variable?)

drivers/pktmem.c
```diff
@@ -17,15 +17,24 @@
...
+#if defined(__ARM_ARCH_7EM__)
+#define CACHE_LINE_SIZE 32
+#else
 #define CACHE_LINE_SIZE 64
-#define PAGE_SIZE       4096
+#endif
 
 #define TRUE_LRU 0
 
+#define PKT_BUF_SIZE (2048 - CACHE_LINE_SIZE)
```
nalajcie, line 29: - what about `imx-enet.c:151` (`state->mmio->MRBR = ENET_BUFFER_SIZE;  // FIXME: coerce with net_allocPktBuf()`)
- incomplete solution, munmap is called with `PAGE_SIZE` as an argument
- is 2048 - 32 really needed as a packet buf? (just asking whether we could use eg.`3 * 512 - 32`)
- IMHO better solution would be to define eg. `PKT_BUF_ALLOC_SIZE` here (depending on `PAGE_SIZE`) and use it in implementation (instead of ifdefed code)
- please note that `mmap` works on full pages, so it's best to alloc multiplicity of PAGE_SIZE (otherwise remaining portion of the last page will never be used)

nalajcie (changes requested):
pointed out further issues with the code.

:exclamation: Please confirm that the code was tested on DTR platform (change the PR to draft until then).

Please split into more (smaller, atomic) commits. **Every potentially-breaking change needs to be in separate commit - clearly describing WHY it's needed**

Please fix commit messages:
 - be more concrete specifying the module in commit message title - here it will mostly be `drivers/imx-enet` or `drivers` - don't list all the files changed (that's what git is for)
 - for fixup comment - write in commit WHAT you're fixing, not WHICH commit (I need to dig further to learn what has actually changed), add info about the commit fixed in commit message body
 - add newline after commit message body and before the trailers

drivers/imx-enet-regs.h
julianuziemblo: - it's also read and written by the driver using `atomic_store`, so shouldn't it be atomic?

nalajcie: have you checked (`git blame`) WHY it's being written using `atomic_store`? IMHO it needs additional comment (regarding WHY we're using `atomic_store` in certain places) instead of type change.

Not sure if `atomic_store` is really needed there, but I assume that it came up in testing.

drivers/ephy.c
```diff
@@ -383,48 +435,74 @@ int ephy_init(eth_phy_state_t *phy, char *conf, uint8_t board_rev, link_state_cb
...
+		err = beginthread(ephy_link_thread, 6, phy->th_stack, sizeof(phy->th_stack), phy);
 		if (err != 0) {
 			gpio_set(&phy->reset, 1);
 			return err;
 		}
 
 		/* enable link up/down IRQ signal */
 		ephy_reg_write(phy, EPHY_1B_IRQ_CTRL_STATUS, (1 << 8) | (1 << 10));
+
+		if ((phy->irq_gpio.flags & GPIO_INVERTED) == 0) {
```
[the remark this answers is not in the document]
julianuziemblo: it's about setting the ephy registers based on provided config, shouldn't it be here? Maybe I should add a function `gpio_inverted` like with `gpio_valid`?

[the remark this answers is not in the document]
nalajcie: why do you want to reconfigure the MACPHY instead of waiting for high / low level in `gpio_wait` depending of the  `GPIO_INVERTED` flag (as it's done for imx6ull?)

drivers/imx-enet.c
```diff
@@ -157,25 +174,57 @@ static void enet_start(enet_state_t *state)
...
-#else
-	state->mmio->TACC = 0;
+		ENET_TACC_SHIFT16 |
+#endif
+		0x0;
+	/* TACC */
+
+	state->mmio->TCR =
+#if ENET_SET_MAC_ON_TX
+		ENET_TCR_ADDINS |
```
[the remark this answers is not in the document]
julianuziemblo: Yes, it works as expected - sets the MAC address in the frame on TX. Should it maybe be a part of another commit or is it not needed at all in your opinion?

drivers/imx-enet.c
```diff
@@ -584,19 +653,23 @@ static uint16_t enet_mdioIO(enet_state_t *state, unsigned addr, unsigned reg, un
 
 #if !ENET_MDC_ALWAYS_ON
 	state->mmio->MSCR = state->mscr;
+	usleep(100); /* NOTE: needed for MDC to adjust to the change */
```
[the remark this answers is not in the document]
julianuziemblo: MDIO doesn't work without it in MDC-not-always-on mode. Maybe we can consider checking for it with e.g. read frames?

drivers/imx-enet.c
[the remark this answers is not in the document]
nalajcie: I just want to avoid adding untested code paths (especially as a part of the commit which should add only a port for `rt106x` as it's disabled for this target).

Did you enable this feature, comment out the parts in lwip stack setting the ETH hw address (or purposely write incorrect data here) - probably in `ethernet_output` and verify on other device (eg. using Wireshark) that the resulting packets have the correct ETH address? Please confirm (the same applies for all other extra features you've added).

Regarding if it's needed - currently there is no graceful way to use it in `lwip` (as per testing path - you would need to manually change the `lwip` code - you should already know that if you have tested it) - I don't expect this feature to be enabled/used soon. It might be useful if/when we would port some other TCP/IP stack so I'm not against adding it if it works.

drivers/imx-enet.c
```diff
@@ -704,24 +815,130 @@ static int enet_initMDIO(enet_state_t *state)
...
-	err = platformctl_seq(&pctl_enet_clock, 1);
-	if (err < 0) {
-		enet_printf(state, "Can't enable ENET clock\n");
+	if (state->dev_phys_addr != ENET_ADDR_ENET1) {
+		enet_warnUnsupportedDeviceAddr(state);
+		return -1;
 	}
 
+	/* RM 14.8.14 */
+#define REG(addr) (*((volatile uint32_t *)(addr)))
```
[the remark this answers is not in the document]
julianuziemblo: there is no public `platformctl` for this. ENET clock is deinitialized in `phoenix-rtos-kernel/hal/(...)/imxrt10xx` - [595](https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/595) fixes this, so it would need to be merged before this

drivers/imx-enet.c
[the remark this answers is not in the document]
julianuziemblo: ok, it will be added later if needed then

drivers/ephy.c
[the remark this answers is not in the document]
julianuziemblo: I see what you mean

drivers/imx-enet.c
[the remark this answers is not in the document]
julianuziemblo: Okay, after some further testing, the MDC-not-always-on code path doesn't seem to be working reliably on neither imxrt1064 nor imx6ull, so a comment should be added or the functionality removed alltogether. Also I don't think this was tested, because it doesn't seem to be working even on [b1ab9f6](https://github.com/phoenix-rtos/phoenix-rtos-lwip/commit/b1ab9f6aa6fa37da945fe368c50334e2f8b71458).

drivers/imx-enet-regs.h
julianuziemblo: ok, I see it's for memory synchronization, will restore the type changed and add an appropriate comment

nalajcie: exactly :+1:

drivers/pktmem.c
julianuziemblo: 1. should I extract the `PKT_BUF_SIZE` to `drivers/pktmem.h` and get this value in `drivers/imx-enet.c` from there?
2. good point
3. `3 * 512 - 32` < 1518, which is the max ethernet frame size, though it's just enough without the optimisation
4. yes

drivers/ephy.c
```diff
@@ -383,48 +435,74 @@ int ephy_init(eth_phy_state_t *phy, char *conf, uint8_t board_rev, link_state_cb
...
+	}
 
 	phy->link_state_callback = cb;
 	phy->link_state_callback_arg = cb_arg;
-	ephy_show_link_state(phy);
+	ephy_setLinkState(phy);
 
 	if (gpio_valid(&phy->irq_gpio)) {
-		err = beginthread(ephy_link_thread, 0, phy->th_stack, sizeof(phy->th_stack), phy);
+		err = beginthread(ephy_link_thread, 6, phy->th_stack, sizeof(phy->th_stack), phy);
```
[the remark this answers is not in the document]
julianuziemblo: after further testing - you're right, it's not needed

drivers/pktmem.c
nalajcie: - IMHO for small targets it's better not to waste memory over minuscule performance improvement due to cache usage
- putting `PKT_BUF_SIZE` in header is a bad idea as the value is not const (it's platform-dependant, actual computation method is implementation detail and it should be hidden from external API). Maybe provide function returning the buffer size instead?

drivers/imxrt106x-gpio.c
```diff
@@ -0,0 +1,232 @@
...
+#include <stdio.h>
+#include <string.h>
+#include <sys/msg.h>
+#include <unistd.h>
+#include <errno.h>
+#include <sys/time.h>
+
+#include "gpio.h"
+#include "imxrt-multi.h"
+#include "lwipopts.h"
```
ziemleszcz, line 23: Probably unnecessary, avoid including unused headers.

drivers/imx-enet.c
```diff
@@ -558,11 +602,27 @@ static int enet_mdioSetup(void *arg, unsigned max_khz, unsigned min_hold_ns, uns
 }
 
 
-static void enet_mdioWait(enet_state_t *state)
+static void enet_mdioWait(enet_state_t *state, time_t timeout)
 {
-	// FIXME: timeout
-	while ((state->mmio->EIR & ENET_IRQ_MII) == 0)
-		/* relax */;
+#define _MDIO_WAIT_CHECK ((state->mmio->EIR & ENET_IRQ_MII) == 0)
```
ziemleszcz, line 608: Unnecessary define, this condition can be written directly.

drivers/imx-enet.c
```diff
@@ -558,11 +602,27 @@ static int enet_mdioSetup(void *arg, unsigned max_khz, unsigned min_hold_ns, uns
...
+	if (timeout != 0) {
+		time_t now, when;
+		gettime(&now, NULL);
+		when = now + timeout;
+
+		while (_MDIO_WAIT_CHECK) {
+			gettime(&now, NULL);
```
ziemleszcz, line 615: Please add small `sleep` here.

drivers/imx-enet.c
```diff
@@ -200,12 +236,13 @@ static void enet_start(enet_state_t *state)
...
 {
-	volatile uint32_t *va = physmmap(0x21bc000, 0x1000);
+	volatile uint32_t *va = physmmap(OCOTP_MEMORY_ADDR, 0x1000);
 
 	if (va == MAP_FAILED) {
 		return -ENOMEM;
 	}
 
+	// FIXME: extract otp memory logic
 	buf[0] = va[0x100 + 4 * 0x22];
```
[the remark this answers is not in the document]
julianuziemblo: according to @nalajcie (https://github.com/phoenix-rtos/phoenix-rtos-lwip/pull/96#discussion_r1742269445), this would be less readable (if I read that comment correctly)

drivers/imx-enet.c
```diff
@@ -558,11 +602,27 @@ static int enet_mdioSetup(void *arg, unsigned max_khz, unsigned min_hold_ns, uns
 }
 
 
-static void enet_mdioWait(enet_state_t *state)
+static void enet_mdioWait(enet_state_t *state, time_t timeout)
```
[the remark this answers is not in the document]
julianuziemblo: this function is only called inside `uint16_t enet_mdioIO`, which is designed as though it "cannot fail". We could return from the latter with a value like 0xffff, but it is a valid register value nontheless. Changing this would require change to the whole MDIO api, I suppose (`mdio_read` and `mdio_write` functions from `drivers/netif-driver.h`).

drivers/imxrt106x-gpio.c
```diff
@@ -0,0 +1,258 @@
...
+	if (*endp != ':' || gp->pin > GPIO_MAX_PIN) {
+		return -EINVAL;
+	}
+
+	arg = endp + 1;
+	if (strlen(arg) < 10) {
+		return -EINVAL;
+	}
+
+	res = strtoul(arg + sizeof(GPIO_PORT_PREFIX) - 1, NULL, 10);
```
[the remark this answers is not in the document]
julianuziemblo: do we want to support paths like `/dev/gpio0x1`?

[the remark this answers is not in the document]
ziemleszcz: Ah, I see now that `res` is part of the filename. My mistake!

drivers/imxrt106x-gpio.c
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	return gpio_setPin(gp, active);
+}
+
+
+int gpio_get(const gpio_info_t *gp)
+{
+	if (!gpio_valid(gp)) {
+		return 0;
```
ziemleszcz, line 143: return -EINVAL

drivers/imx6ull-gpio.c
```diff
@@ -99,3 +106,203 @@ int imx6ull_gpio_config(const char *name, uint32_t mask, unsigned flags)
...
+}
+
+
+int gpio_get(const gpio_info_t *gp)
+{
+	uint32_t data[1];
+	int err;
+
+	if (!gpio_valid(gp)) {
+		return 0;
```
ziemleszcz, line 149: return -EINVAL

drivers/imx6ull-gpio.c
```diff
@@ -99,3 +106,203 @@ int imx6ull_gpio_config(const char *name, uint32_t mask, unsigned flags)
...
+	}
+
+	err = read(gp->fd, data, sizeof(data));
+	if (err < 0) {
+		return -errno;
+	}
+	if (err != sizeof(data)) {
+		return 0;
```
ziemleszcz, line 157: return -EIO

drivers/imx-enet.c
```diff
@@ -404,7 +404,7 @@ static void enet_fillRxDesc(const net_bufdesc_ring_t *ring, size_t i, addr_t pa,
 #if ENET_USE_ENHANCED_DESCRIPTORS
 	desc->yflags = ENET_RXDY_INT;
 #endif
-	atomic_store(&desc->flags, ENET_DESC_RDY | wrap);
+	atomic_store(&desc->flags, ENET_DESC_RDY | wrap); /* NOTE: memory sync */
```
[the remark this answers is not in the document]
julianuziemblo: we don't want `flags` to be atomic, because they are written to by hardware. We use `atomic_store` for memory sync only. Should the comment be more specific?

discussion on this exact topic: https://github.com/phoenix-rtos/phoenix-rtos-lwip/pull/96#discussion_r1771282235

drivers/imx6ull-gpio.c
```diff
@@ -100,3 +108,206 @@ int imx6ull_gpio_config(const char *name, uint32_t mask, unsigned flags)
...
+	if (err < 0) {
+		return err;
+	}
+
+	snprintf(buf, sizeof(buf), "%s/port", arg);
+	gp->fd = open(buf, O_RDWR);
+	if (gp->fd < 0) {
+		err = -errno;
+		printf("gpio: can't open '%s': %s (%d)\n", buf, strerror(errno), -errno);
+		if (fd >= 0) {
```
ziemleszcz, line 265: `fd == -1` for sure at this point.

drivers/imx6ull-gpio.c
```diff
@@ -100,3 +108,206 @@ int imx6ull_gpio_config(const char *name, uint32_t mask, unsigned flags)
...
+	}
+
+	if ((flags & GPIO_OUTPUT) != 0) {
+		err = gpio_do_set(gp, !!(flags & GPIO_ACTIVE) ^ !!(flags & GPIO_INVERTED));
+		if (err != 0) {
+			printf("gpio: WARN: can't pre-set pin: %s (%d; fd = %d)\n", strerror(-err), err, gp->fd);
+		}
+	}
+
+	fd = gp->fd;
```
ziemleszcz, line 278: Maybe add comment "Save gpio port file descriptor".

drivers/imx6ull-gpio.c
```diff
@@ -100,3 +108,206 @@ int imx6ull_gpio_config(const char *name, uint32_t mask, unsigned flags)
...
+	close(gp->fd);
+
+	if (err != 0) {
+		if (fd >= 0) {
+			close(fd);
+		}
+		return err;
+	}
+
+	gp->fd = fd;
```
ziemleszcz, line 305: Maybe add comment "Restore gpio port file descriptor".

drivers/imx6ull-gpio.c
```diff
@@ -100,3 +108,206 @@ int imx6ull_gpio_config(const char *name, uint32_t mask, unsigned flags)
...
+
+	return err;
+}
+
+
+int gpio_init(gpio_info_t *gp, const char *arg, unsigned flags)
+{
+	char buf[64];
+	char *endp;
+	int err, fd = -1;
```
ziemleszcz, line 231: There is no need to initialize `fd`.

drivers/imx6ull-gpio.c
```diff
@@ -100,3 +108,206 @@ int imx6ull_gpio_config(const char *name, uint32_t mask, unsigned flags)
...
+	}
+
+	fd = gp->fd;
+
+	snprintf(buf, sizeof(buf), "%s/dir", arg);
+	gp->fd = open(buf, O_RDWR);
+	if (gp->fd < 0) {
+		err = -errno;
+		printf("gpio: can't open '%s': %s (%d)\n", buf, strerror(-err), err);
+		if (fd >= 0) {
```
ziemleszcz, line 285: `fd >= 0` for sure at this point.

drivers/imx6ull-gpio.c
```diff
@@ -100,3 +108,206 @@ int imx6ull_gpio_config(const char *name, uint32_t mask, unsigned flags)
...
+		}
+		return err;
+	}
+
+	err = gpio_do_set(gp, flags & GPIO_OUTPUT);
+	if (err != 0) {
+		printf("gpio: can't configure pin direction: %s (%d; fd = %d)\n", strerror(-err), err, gp->fd);
+	}
+
+	close(gp->fd);
```
ziemleszcz, line 296: Move the code from lines 300-307 inside `if` statement above. Notice that `fd >= 0` for sure at this point.

drivers/imx-enet.c
```diff
@@ -172,9 +210,9 @@ static void enet_start(enet_state_t *state)
 #else
 	state->mmio->TACC = 0;
 #endif
-	/* TACC */
```
ziemleszcz: I find the commenting strategy in this function unclear.

drivers/imx-enet.c
```diff
@@ -33,31 +31,61 @@
...
+#if defined(__CPU_IMXRT106X)
+
+#include <phoenix/arch/armv7m/imxrt/10xx/imxrt10xx.h>
+
+#define ENET_ADDR_ENET1   (0x402D8000)
+#define OCOTP_MEMORY_ADDR (0x401F4000)
+
+#define ENET_CLK_KHZ (132000)
+
+#define ENET_RX_RING_SIZE 32
```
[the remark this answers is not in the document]
julianuziemblo: Yes, 64 was too much on imxrt1064 and lead to out-of-memory bugs

drivers/imx-enet.c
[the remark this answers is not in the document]
ziemleszcz: This is a poor API design on the lwIP side. Linux MDIO drivers return `ETIMEDOUT` on timeout. I still believe `enet_mdioWait` should return a status, and `enet_mdioIO` should at least print a warning on timeout (reusing the debug message from `enet_mdioWait`). We cannot assume that MDIO will never fail, future code changes or support for new chips could easily break it.

drivers/ephy.c
```diff
@@ -99,67 +124,73 @@ static uint32_t ephy_show_id(eth_phy_state_t *phy)
...
-	bstat = ephy_reg_read(phy, 0x01);
-	bstat = ephy_reg_read(phy, 0x01);
-	adv = ephy_reg_read(phy, 0x04);
-	lpa = ephy_reg_read(phy, 0x05);
-	pc1 = ephy_reg_read(phy, 0x1e);
-	pc2 = ephy_reg_read(phy, 0x1f);
-	speed = ephy_link_speed(phy, &full_duplex);
+	bctl = ephy_reg_read(phy, EPHY_00_BMCR);
+	bstat = ephy_reg_read(phy, EPHY_01_BMSR);
+	bstat = ephy_reg_read(phy, EPHY_01_BMSR);
```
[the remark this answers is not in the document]
julianuziemblo: Seems like I confused KSZ8081 with RTL8201, the latter really needs 2 reads, but I tested KSZ8081 and it seems like it doesn't need the extra read. I'll remove it.

drivers/imx-enet.c
```diff
@@ -0,0 +1,1435 @@
...
+	enet_mdioWrite,
+};
+
+
+static int platformctl_seq(const platformctl_t *pctl, size_t n)
+{
+	int err = 0;
+
+	for (; n--; ++pctl) {
+		err = platformctl((platformctl_t *)pctl);
```
ziemleszcz, line 728: This leads to undefined behavior. Please declare a local `platformctl_t` variable and copy the input into it.

drivers/imx-enet.c
```diff
@@ -0,0 +1,1435 @@
...
+}
+
+
+/*
+ * Enable clock for the required ENET module
+ */
+static int enet_clockEnable(enet_state_t *state)
+{
+#if defined(__CPU_IMXRT106X)
+
```
ziemleszcz, line 825: To clarify, my suggestion to add empty lines was intended for a specific location (ENET registers). In this case, the code looks fine without additional spacing.

drivers/imx6ull-gpio.c
```diff
@@ -71,27 +84,226 @@ static int imx6ull_set_pin_conf(unsigned mux, unsigned pad, unsigned flags)
...
-			err = imx6ull_set_pin_conf(pinc->mux0 + ov, pinc->pad0 + ov, flags);
+			err = set_pin_conf(pinc->mux0 + ov, pinc->pad0 + ov, flags);
+		}
+	}
+
+	return err;
+}
+
+
+static int gpio_do_set(const gpio_info_t *gp, int val)
```
ziemleszcz, line 115: `gpio_doSet`

drivers/imx6ull-gpio.c
```diff
@@ -22,26 +33,26 @@ struct pin_conf {
...
-	{ 0, }
+	{ 0 }
 };
 
-static const struct pin_conf *const imx_gpio_pin[5] = {  pin_conf1,  pin_conf0,  pin_conf0,  pin_conf0,  pin_conf5 };
+static const struct pin_conf *const imx_gpio_pin[5] = { pin_conf1, pin_conf0, pin_conf0, pin_conf0, pin_conf5 };
 
 
-static int imx6ull_set_pin_conf(unsigned mux, unsigned pad, unsigned flags)
+static int set_pin_conf(unsigned mux, unsigned pad, unsigned flags)
```
ziemleszcz, line 55: `gpio_setPinConf`

drivers/imx-enet.c
julianuziemblo: ah, okay, will revise those other places then

drivers/bdring.c
```diff
@@ -68,40 +70,45 @@ int net_initRings(net_bufdesc_ring_t *rings, const size_t *sizes, size_t nrings,
...
+		}
 	}
 
 	bufp = calloc(nb, sizeof(*bufp));
-	if (!bufp)
+	if (bufp == NULL) {
 		return -ENOMEM;
+	}
 
 	p = dmammap(sz);
```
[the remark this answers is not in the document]
ziemleszcz: On second thought, it might be better to allocate all the descriptors in one continuous block to save memory, as long as they fit in a single page.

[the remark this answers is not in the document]
julianuziemblo: so you want a check to be added to test if they all fit on 1 page, and if not, make them not continuous?

drivers/imx-enet.c
julianuziemblo: I changed it at the top of the file, but I think it's more readable this way in the functions that use platformctl

drivers/pktmem.c
```diff
@@ -17,28 +17,55 @@
...
 
 
-#define PKT_BUF_SIZE       (2048 - CACHE_LINE_SIZE)
-#define PKT_BUF_CNT        (size_t)((PAGE_SIZE - CACHE_LINE_SIZE) / PKT_BUF_SIZE)
-#define PKT_BUF_IDX        11 /* log2(PAGE_SIZE) - ceil(log2(PKT_BUF_CNT)) */
-#define PKT_BUF_CACHE_SIZE 16
+#if PAGE_SIZE < PKT_BUF_SIZE
+#define PKT_BUF_CNT              1
+#define PKT_BUF_WHICH(p)         0
+#define PKT_BUF_PAGE_HEAD_OFFSET (PKT_BUF_ALLOC_SIZE - CACHE_LINE_SIZE)
```
ziemleszcz, line 37: Does this mean that `_buf_page_head` is positioned at the end of the last page in this case? Why?

drivers/pktmem.c
```diff
@@ -17,28 +17,55 @@
 #include <string.h>
 
 
-#define CACHE_LINE_SIZE 64
-#define PAGE_SIZE       4096
+// FIXME: transfer CACHE_LINE_SIZE to appropriate file with platform-specific definitions
+#if defined(__ARM_ARCH_7EM__)
+#define CACHE_LINE_SIZE    32
+#define PKT_BUF_ALLOC_SIZE (3 * PAGE_SIZE)
+#define PKT_BUF_SIZE       (PKT_BUF_ALLOC_SIZE - 12 /* sizeof(buf_page_head_t) (without alignment) */)
```
ziemleszcz, line 24: Why not `PKT_BUF_ALLOC_SIZE - CACHE_LINE_SIZE`?

drivers/pktmem.c
```diff
@@ -17,28 +17,55 @@
...
+#endif
 
 #define TRUE_LRU 0
 
 
-#define PKT_BUF_SIZE       (2048 - CACHE_LINE_SIZE)
-#define PKT_BUF_CNT        (size_t)((PAGE_SIZE - CACHE_LINE_SIZE) / PKT_BUF_SIZE)
-#define PKT_BUF_IDX        11 /* log2(PAGE_SIZE) - ceil(log2(PKT_BUF_CNT)) */
-#define PKT_BUF_CACHE_SIZE 16
+#if PAGE_SIZE < PKT_BUF_SIZE
```
ziemleszcz, line 34: Why not merge this `#if` with the `#if defined(__ARM_ARCH_7EM__)` above?

drivers/pktmem.c
```diff
@@ -17,28 +17,55 @@
...
+#define PKT_BUF_CNT              1
+#define PKT_BUF_WHICH(p)         0
+#define PKT_BUF_PAGE_HEAD_OFFSET (PKT_BUF_ALLOC_SIZE - CACHE_LINE_SIZE)
+#define PKT_BUF_PACKET_OFFSET    0
 
+typedef struct _buf_page_head {
+	uint32_t rsvd[5]; /* NOTE: aligned access */
```
ziemleszcz, line 41: Could you explain why this is necessary?

drivers/pktmem.c
```diff
@@ -17,28 +17,55 @@
...
+	uint32_t free_mask;
+	struct _buf_page_head *next;
+	struct _buf_page_head *prev;
+} buf_page_head_t;
+#else
+#define PKT_BUF_CNT              (size_t)((PAGE_SIZE - CACHE_LINE_SIZE) / PKT_BUF_SIZE)
```
ziemleszcz, line 47: `(PKT_BUF_ALLOC_SIZE - CACHE_LINE_SIZE) / PKT_BUF_SIZE`?

drivers/pktmem.c
julianuziemblo: Yes, it's that way because of space and alignment restrictions on smaller targets - not to waste memory.
Specifically: on imxrt106x, the page size is 512, and Ethernet MTU is typically 1500 + 18B of header. We can fit 1 packet on 3 pages, but the address of the packet we give to hardware must be on a 32B-boundary. That's why we can place the page head at the end of the buffer to avoid this problem.
I will position it at the end for all cases to keep consistency.

drivers/pktmem.c
julianuziemblo: After testing - it seems that it is not necessary, will remove it.

drivers/pktmem.c
julianuziemblo: it avoids space problems on smaller targets, as mentioned in the response to previous comment.

drivers/pktmem.c
```diff
@@ -140,14 +156,67 @@ static ssize_t net_allocPktBuf(void **bufp)
...
+	pkt_buf_opts.alloc_sz = PAGE_SIZE * ((pkt_buf_max_sz / PAGE_SIZE) + 1);
+
+	if (pkt_buf_max_sz + CACHE_LINE_SIZE <= pkt_buf_opts.alloc_sz) {
+		/* Cache line optimisation */
+		page_head_sz = CACHE_LINE_SIZE;
+		pkt_buf_opts.pkt_cnt = PAGE_SIZE / (pkt_buf_max_sz + page_head_sz);
+		pkt_buf_opts.which = net_whichFromAddr;
+	}
+	else {
+		/* No memory - skip the optimisation */
```
ziemleszcz, line 200: This should never happen, DMA can now overwrite `buf_page_head_t` header.

drivers/pktmem.c
```diff
@@ -140,14 +156,67 @@ static ssize_t net_allocPktBuf(void **bufp)
...
+	}
+
+	size_t page_head_sz;
+
+	pkt_buf_opts.alloc_sz = PAGE_SIZE * ((pkt_buf_max_sz / PAGE_SIZE) + 1);
+
+	if (pkt_buf_max_sz + CACHE_LINE_SIZE <= pkt_buf_opts.alloc_sz) {
+		/* Cache line optimisation */
+		page_head_sz = CACHE_LINE_SIZE;
+		pkt_buf_opts.pkt_cnt = PAGE_SIZE / (pkt_buf_max_sz + page_head_sz);
```
ziemleszcz, line 196: Probably it should be `(pkt_buf_opts.alloc_sz - CACHE_LINE_SIZE) / pkt_buf_max_sz_aligned` where `pkt_buf_max_sz_aligned` is `pkt_buf_max_sz` DMA aligned.

drivers/pktmem.c
```diff
@@ -140,14 +156,67 @@ static ssize_t net_allocPktBuf(void **bufp)
 
 	SYS_ARCH_UNPROTECT(old_level);
 
-	*bufp = (void *)ph + CACHE_LINE_SIZE + i * PKT_BUF_SIZE;
+	*bufp = (void *)ph - pkt_buf_opts.page_head_offset + (i * (pkt_buf_opts.alloc_sz / pkt_buf_opts.pkt_cnt));
```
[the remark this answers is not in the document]
julianuziemblo: `pkt_sz` can be any number (1518 for enet, 1524 for RTL), so we cannot assume that. `pkt_buf_opts.alloc_sz / pkt_buf_opts.pkt_cnt` works for every case.

drivers/pktmem.c
ziemleszcz: `pkt_buf_max_sz + CACHE_LINE_SIZE > pkt_buf_opts.alloc_sz`

ziemleszcz: `pkt_buf_opts.alloc_sz` should be large enough to hold at least one packet buffer (`pkt_buf_max_sz`) and the header (`CACHE_LINE_SIZE`).

drivers/pktmem.c
```diff
@@ -140,14 +156,67 @@ static ssize_t net_allocPktBuf(void **bufp)
...
+		pkt_buf_opts.which = net_whichFromAddr;
+	}
+	else {
+		/* No memory - skip the optimisation */
+		page_head_sz = sizeof(buf_page_head_t);
+		pkt_buf_opts.pkt_cnt = 1;
+		pkt_buf_opts.which = net_whichAlwaysZero;
+	}
+
+	pkt_buf_opts.pkt_sz = (pkt_buf_opts.alloc_sz / pkt_buf_opts.pkt_cnt) - page_head_sz;
```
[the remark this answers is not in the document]
julianuziemblo: I dont think that's true, the packet buffer can be as big as we can afford, up to the page head at the end of the buffer

julianuziemblo:
/gemini are there any problems with pointer arithmetic in pktmem.c introduced in commit [bdd6223](https://github.com/phoenix-rtos/phoenix-rtos-lwip/pull/96/commits/bdd6223b096cd2c2189a26954940427a2ae9aefe) for target with PAGE_SIZE=512?

drivers/ephy.c
```diff
@@ -22,223 +22,276 @@
...
+	}
 
-	err = gpio_init(pin, cfg, flags);
-	if (err) {
-		printf("ephy: %s bad pin info: %s (%d)\n", pfx, strerror(err), err);
+	err = gpio_init(gp, cfg, flags);
+	if (err != 0) {
+		printf("lwip: ephy: %s bad pin info: %s (%d)\n", pfx, strerror(-err), err);
 		return cfg - pfx_len;
 	}
```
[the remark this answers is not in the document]
agkaminski: lol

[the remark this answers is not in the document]
nalajcie: /gemini but it is used - exactly in the code snipped you've provided?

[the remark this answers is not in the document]
ziemleszcz: /gemini Gemini, don’t apologize. These guys are just being harsh!

drivers/bdring.c
```diff
@@ -41,67 +42,72 @@
  */
 
 
-static int is_power_of_2(size_t n)
+static bool net_isPowerOf2(size_t n)
 {
-	return !(n & (n - 1));
+	return (n & (n - 1)) == 0;
```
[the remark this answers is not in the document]
julianuziemblo: @gemini-code-assist how is returning n == 0 a substitute for current logic, which checks if n is a power of 2?

drivers/pktmem.c
```diff
@@ -15,80 +15,160 @@
...
+int net_initPktMem(size_t pkt_buf_max_sz)
+{
+	if (net_isPktMemInitialized()) {
+		return 0;
+	}
+
+	if (pkt_buf_max_sz == 0) {
+		return -EINVAL;
+	}
+
```
ziemleszcz, line 116: Please return `-EINVAL` if `sizeof(pktmem_info_t) % __alignof__(pktmem_info_t) != 0` to ensure that `pktmem_info_t` structure is properly aligned.

drivers/pktmem.c
```diff
@@ -15,80 +15,159 @@
...
+	const size_t pktmem_info_sz = sizeof(pktmem_info_t);
+
+	pktmem_opts.alloc_sz = PAGE_SIZE * ((pkt_buf_max_sz + pktmem_info_sz + PAGE_SIZE - 1) / PAGE_SIZE);
+	pktmem_opts.pkt_buf_sz = pkt_buf_max_sz;
+	pktmem_opts.pkt_buf_cnt = 1 + (pktmem_opts.alloc_sz - pkt_buf_max_sz - pktmem_info_sz) / pkt_buf_max_sz;
+	pktmem_opts.pktmem_info_offset = pktmem_opts.alloc_sz - pktmem_info_sz;
+	pktmem_opts.free_mask_full = (1 << pktmem_opts.pkt_buf_cnt) - 1;
+
```
ziemleszcz, line 124: Please return `-EINVAL` or limit `pktmem_opts.pkt_buf_cnt` if `pktmem_opts.pkt_buf_cnt > sizeof(pktmem_opts.free_mask_full) * 8`.

drivers/pktmem.c
```diff
@@ -97,86 +176,102 @@ static void net_freeDMAPbuf(struct pbuf *p)
...
 	unsigned i;
 
 	SYS_ARCH_PROTECT(old_level);
-	if (pkt_bufs_free == 0) {
+	if (pkt_bufs_free == 0 || pktmem_lru.next == &pktmem_lru) {
 		SYS_ARCH_UNPROTECT(old_level);
 
-		ph = dmammap(PAGE_SIZE);
-		if (ph == NULL) {
+		info = dmammap(pktmem_opts.alloc_sz);
```
ziemleszcz, line 197: Consider assigning the result to a new variable, to simplify the calculation at lines 225-226.

drivers/pktmem.c
```diff
@@ -97,86 +176,102 @@ static void net_freeDMAPbuf(struct pbuf *p)
...
+		info = pktmem_lru.next;
 	}
 
-	i = __builtin_ctz(ph->free_mask);
 	--pkt_bufs_free;
-	ph->free_mask &= ~(1 << i);
-	if (ph->free_mask == 0) {
-		net_listDel(ph);
+
+	i = __builtin_ctz(info->free_mask);
```
ziemleszcz, line 217: You could optimize the new pktmem case by moving this code into `else` block. In the case of a newly allocated pktmem with only one packet buffer, we don't need to add and remove it from the list.

drivers/pktmem.c
```diff
@@ -15,80 +15,159 @@
...
+
+	return 0;
+}
+
+
 static void net_freePktBuf(void *p)
 {
-	buf_page_head_t *ph = (void *)((uintptr_t)p & ~(PAGE_SIZE - 1));
-	unsigned which = ((size_t)p & (PAGE_SIZE - 1)) >> PKT_BUF_IDX;
+	pktmem_info_t *info = (void *)(((uintptr_t)p & ~(PAGE_SIZE - 1)) + pktmem_opts.pktmem_info_offset);
```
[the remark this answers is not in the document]
julianuziemblo: I think this is a comment to the wrong commit, this is fixed in the new version

drivers/pktmem.h
```diff
@@ -12,11 +12,13 @@
 #define NET_PKTMEM_H_
 
 #include <stdint.h>
+#include <unistd.h>
 
 
 struct pbuf;
```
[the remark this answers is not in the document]
julianuziemblo: We need `unistd.h` for `size_t` and `addr_t` types here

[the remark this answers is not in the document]
ziemleszcz: Remove `struct pbuf`.

drivers/pktmem.c
```diff
@@ -15,80 +15,160 @@
...
+static inline bool net_isPktMemInitialized(void)
+{
+	return pktmem_opts.alloc_sz != 0;
+}
+
+
+int net_initPktMem(size_t pkt_buf_max_sz)
+{
+	if (net_isPktMemInitialized()) {
+		return 0;
```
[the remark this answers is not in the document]
julianuziemblo: ahh yes, silly oversight...

ziemleszcz (approved):
LGTM

PR #109, opened 2025-07-24
drivers/greth: add support for ethernet in gr740

atlomak:
## Description

Add ethernet driver for `GR740` and add support for MAC interrupts in `ephy.c`. 

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `sparcv8leon-gr740`, `imxrt106x`, `imxrt117x`.

## Checklist:

## Special treatment

### DIFF
--- a/drivers/ephy.c
+++ b/drivers/ephy.c
@@ -269,7 +269,7 @@ static inline int ephy_ksz8081rnx_linkSpeed(const eth_phy_state_t *phy, int *ful
 	}
 
 	if (full_duplex != NULL) {
-		*full_duplex = !!(pc1 & (1u << 2));
+		*full_duplex = ((pc1 & (1u << 2)) != 0) ? 1 : 0;
 	}
 
 	return ((pc1 & 0x1) != 0) ? 10 : 100;
@@ -309,7 +309,7 @@ static inline int ephy_rtl8201fi_linkSpeed(const eth_phy_state_t *phy, int *full
 	uint16_t bmcr = ephy_regRead(phy, EPHY_COMMON_00_BMCR);
 
 	if (full_duplex != NULL) {
-		*full_duplex = !!(bmcr & (1 << 8));
+		*full_duplex = ((bmcr & (1 << 8)) != 0) ? 1 : 0;
 	}
 
 	return ((bmcr & (1 << 13)) == 0) ? 10 : 100;
@@ -322,7 +322,7 @@ static inline int ephy_rtl8211fdi_linkSpeed(const eth_phy_state_t *phy, int *ful
 	uint16_t speed;
 
 	if (full_duplex != NULL) {
-		*full_duplex = !!(physr & (1 << 3));
+		*full_duplex = ((physr & (1 << 3)) != 0) ? 1 : 0;
 	}
 
 	speed = (physr >> 4) & 0x3;


_targets/Makefile.sparcv8leon-gr740
```diff
@@ -0,0 +1,12 @@
+#
+# Makefile for Phoenix-RTOS 3 LwIP
+#
+# sparcv8leon/gr740-mini target
+#
+# Copyright 2020 Phoenix Systems
```
lukileczo, line 6:
```suggestion
# Copyright 2025 Phoenix Systems
```

include/default-opts/lwipopts.h
```diff
@@ -29,6 +29,12 @@
 #define MSCHAP_SUPPORT 1
 #define LWIP_TIMEVAL_PRIVATE 0
 
+#ifdef __CPU_GR740
+#define CHECKSUM_GEN_TCP 0
+#define CHECKSUM_GEN_UDP 0
+#define CHECKSUM_GEN_IP  0
+#endif
```
lukileczo, line 36: The default options should remain unchanged, please add `lwipopts.h` to `phoenix-rtos-project/_projects/sparcv8leon-gr740-mini/lwip/` directory and set `export LWIPOPTS_DIR="$(pwd)/_projects/sparcv8leon-gr740-mini/lwip"` in the `build.project` for gr740-mini target - see e.g. https://github.com/phoenix-rtos/phoenix-rtos-project/blob/9d39cf340b7678e584882d2ac3b5f8f99e700b0e/_projects/ia32-generic-qemu/build.project#L14 for reference

drivers/greth.c
```diff
@@ -0,0 +1,564 @@
...
+	atomic_thread_fence(memory_order_seq_cst);
+	desc->flags = GRETH_DESC_EN | GRETH_DESC_IE | wrap;
+	atomic_thread_fence(memory_order_seq_cst);
+}
+
+
+static int greth_nextTxDone(const net_bufdesc_ring_t *ring, size_t i)
+{
+	volatile greth_buf_desc_t *desc = (volatile greth_buf_desc_t *)ring->ring + i;
+	return !(desc->flags & GRETH_DESC_EN);
```
lukileczo, line 280: MISRA
```suggestion
	return ((desc->flags & GRETH_DESC_EN) == 0u) ? 1 : 0;
```

drivers/greth.c
```diff
@@ -0,0 +1,564 @@
...
+}
+
+
+static err_t greth_netifOutput(struct netif *netif, struct pbuf *p)
+{
+	greth_state_t *state = netif->state;
+	size_t nf;
+
+	if (ETH_PAD_SIZE != 0) {
+		(void)pbuf_header(p, -ETH_PAD_SIZE); /* drop the padding word */
```
lukileczo, line 343: Maybe comment about why we're doing that
```suggestion
		(void)pbuf_header(p, -ETH_PAD_SIZE); /* drop the padding word (not supported by HW) */
```

drivers/greth.c
```diff
@@ -0,0 +1,564 @@
...
+static inline uint8_t greth_getByte(uint32_t v, int i)
+{
+	return (v >> (i * 8)) & 0xFF;
+}
+
+
+static void greth_readCardMac(greth_state_t *state)
+{
+	uint8_t *mac;
+	mac = (void *)&state->netif->hwaddr;
```
lukileczo, line 366:
```suggestion
	uint8_t *mac = &state->netif->hwaddr;
```

drivers/greth.c
```diff
@@ -0,0 +1,564 @@
...
+static int greth_initDevice(greth_state_t *state, int irq, int mdio)
+{
+	state->mmio = physmmap(state->dev_phys_addr, 0x1000);
+	if (state->mmio == MAP_FAILED) {
+		return -ENOMEM;
+	};
+
+#ifdef GRETH_DISABLE_EDCL
+	int err = greth_disable_edcl(state);
+#endif
```
lukileczo, line 434: Maybe such configuration should reside in the `board_config.h` file for the target?

drivers/greth.c
```diff
@@ -0,0 +1,564 @@
...
+#define greth_debug_printf(state, ...) greth_printf(state, __VA_ARGS__)
+#else
+#define greth_debug_printf(...)
+#endif
+
+
+typedef struct {
+	volatile struct greth_regs *mmio;
+	struct netif *netif;
+	volatile bool drv_exit;
```
lukileczo, line 59:
```suggestion
	atomic_bool drv_exit;
```

drivers/greth.c
```diff
@@ -0,0 +1,564 @@
...
+		return err;
+	}
+
+	err = interrupt(irq, greth_irqHandler, state, state->irq_cond, NULL);
+	if (err != 0) {
+		greth_printf(state, "Couldn't register interrupt handler: %s (%d)", strerror(-err), err);
+		return err;
+	}
+
+	err = beginthread(greth_irqThread, 4, state->irq_stack, sizeof(state->irq_stack), state);
```
lukileczo, line 471: Priority 4 is too low for irq thread, I'd set it to 1
```suggestion
	err = beginthread(greth_irqThread, 1, state->irq_stack, sizeof(state->irq_stack), state);
```

drivers/greth.c
```diff
@@ -0,0 +1,564 @@
...
+	if (err != 0) {
+		greth_printf(state, "Couldn't create resources: %s (%d)", strerror(-err), err);
+		greth_cleanupResources(state);
+		return err;
+	}
+
+	err = interrupt(irq, greth_irqHandler, state, state->irq_cond, NULL);
+	if (err != 0) {
+		greth_printf(state, "Couldn't register interrupt handler: %s (%d)", strerror(-err), err);
+		return err;
```
lukileczo, line 468:
```suggestion
		greth_cleanupResources(state);
		return err;
```

drivers/greth.c
```diff
@@ -0,0 +1,564 @@
...
+	}
+
+	err = beginthread(greth_irqThread, 4, state->irq_stack, sizeof(state->irq_stack), state);
+	if (err != 0) {
+		greth_printf(state, "Couldn't begin interrupt thread: %s (%d)", strerror(-err), err);
+		return err;
```
lukileczo, line 474:
```suggestion
		greth_cleanupResources(state);
		return err;
```

drivers/greth.c
```diff
@@ -0,0 +1,564 @@
...
+	volatile greth_buf_desc_t *desc = (volatile greth_buf_desc_t *)ring->ring + i;
+	uint32_t flags;
+
+	flags = GRETH_DESC_EN;
+	flags |= (i == ring->last) ? GRETH_DESC_WR : 0;
+	flags |= (sz << GRETH_DESC_LEN_SHIFT) & GRETH_DESC_LEN_MASK; /* shift (0) just for consistency  */
+	flags |= ((seg & BDRING_SEG_LAST) != 0) ? 0 : GRETH_DESC_TX_MO;
+	flags |= GRETH_DESC_IE;
+
+#if GRETH_CHECKSUM_OFFLOAD
```
lukileczo, line 295: Is there a situation in which we'd not want to offload it? If not, then I'd remove the `#define`, otherwise please move it to `board_config.h`

drivers/greth.c
```diff
@@ -0,0 +1,564 @@
...
+#define GRETH_MAX_PKT_SZ       1514
+
+typedef greth_desc_t greth_buf_desc_t;
+
+#define GRETH_DEBUG      0
+#define GRETH_MDIO_DEBUG 0
+#if GRETH_DEBUG
+#define greth_debug_printf(state, ...) greth_printf(state, __VA_ARGS__)
+#else
+#define greth_debug_printf(...)
```
lukileczo, line 52: Generally the convention is for all `#define`s to be uppercase. I see that in imx it's defined similarly to how you've done it, but I think we should stick to our coding convention, @agkaminski what's your opinion?

lukileczo:
Please update also readme with instructions for using the new driver

drivers/greth.c
julianuziemblo: if so: it should be also `atomic_store`'d, `atomic_load`'ed and `atomic_init`ialized in driver's init function

drivers/greth.c
```diff
@@ -0,0 +1,564 @@
...
+	}
+	return sz;
+}
+
+
+static int greth_pktRxFinished(const net_bufdesc_ring_t *ring, size_t i)
+{
+	volatile greth_buf_desc_t *desc = (volatile greth_buf_desc_t *)ring->ring + i;
+
+	return !(desc->flags & GRETH_DESC_EN);
```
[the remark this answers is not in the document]
julianuziemblo: ... or change the method's return type to `bool`, as it's already used as such (in a separate commit of course)

also do this in `drivers/imx-enet.c`, `drivers/rtl8139cp.c` with `<module>_nextTxDone` and `<module>_pktRxFinished` functions

drivers/greth.c
julianuziemblo: also: `ETH_PAD_SIZE` is known at compile time, so you should check it with preprocessor

if there's anything similar in other drivers, please fix it also

drivers/ephy.c
```diff
@@ -272,6 +282,34 @@ static inline int ephy_ksz8081rnx_linkSpeed(const eth_phy_state_t *phy, int *ful
...
+{
+	uint16_t st = ephy_regRead(phy, EPHY_COMMON_01_BMSR);
+	/* PHY still in auto-negotiation */
+	if ((st & (1u << 5)) == 0) {
+		return 0;
+	}
+
+	uint16_t pc = ephy_regRead(phy, EPHY_KSZ9031_1F_PHYCR);
+	if (full_duplex != NULL) {
+		*full_duplex = !!(pc & (1u << 3));
```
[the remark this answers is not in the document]
julianuziemblo: change the others ones accordingly (in separate commit)

drivers/greth.c
lukileczo: Not really, the atomic operations are already generated when the variable is marked as such: https://godbolt.org/z/GEaTePY5c
I can't find in the C standard now, but see also https://developers.redhat.com/blog/2016/01/14/toward-a-better-use-of-c11-atomics-part-1

drivers/ephy.h
```diff
@@ -22,12 +22,19 @@ typedef struct {
...
 		ephy_rtl8211fdi } model;
 	unsigned bus;
 	unsigned addr;
 	unsigned reset_hold_time_us;
 	unsigned reset_release_time_us;
+
+#define EPHY_GPIO_INTERRUPT -1
+#define EPHY_MAC_INTERRUPT  -2
+	int irq;
+	uint32_t irq_mask;
```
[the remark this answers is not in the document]
atlomak: good idea

drivers/greth.c
atlomak: ok, changed

drivers/greth.c
```diff
@@ -0,0 +1,562 @@
...
+
+static void greth_cleanupResources(greth_state_t *state)
+{
+	if (state->irq_lock < (handle_t)-1) {
+		resourceDestroy(state->irq_lock);
+	}
+	if (state->irq_cond < (handle_t)-1) {
+		resourceDestroy(state->irq_cond);
+	}
+}
```
julianuziemblo, line 418: check should be `state->irq_lock != (handle_t)-1` and you should explicitly set `state->irq_lock = -1` before creating it (e.g. in `greth_initDevice`), as we don't know what value it holds before initialization

drivers/greth.c
```diff
@@ -0,0 +1,562 @@
...
+	if (err != 0) {
+		return err;
+	}
+	greth_debug_printf(state, "Initialized GRETH Rings");
+
+	if (err != 0) {
+		greth_printf(state, "Couldn't register interrupt handler: %s (%d)", strerror(-err), err);
+		return err;
+	}
+	greth_debug_printf(state, "Interrupt handler initialized successfully");
```
julianuziemblo, line 442: I don't think this should be here

drivers/greth.c
```diff
@@ -0,0 +1,562 @@
...
+	if (err != 0) {
+		greth_printf(state, "Couldn't begin interrupt thread: %s (%d)", strerror(-err), err);
+		greth_cleanupResources(state);
+		return err;
+	}
+
+	greth_readCardMac(state);
+	greth_showCardId(state);
+
+	greth_startGreth(state);
```
julianuziemblo, line 478: nitpick: just `greth_start` should be less cluttered - but it's really just a preference :smile:

drivers/greth.c
```diff
@@ -0,0 +1,562 @@
...
+
+	return err;
+};
+
+
+static int greth_netifInit(struct netif *netif, char *cfg)
+{
+	greth_state_t *state;
+	char *p;
+	int err, irq, mdio = 1;
```
julianuziemblo, line 488: `mdio` should just be a boolean - see my previous comment

drivers/greth.c
```diff
@@ -0,0 +1,562 @@
...
+	return 0;
+}
+
+
+static netif_driver_t greth_drv = {
+	.init = greth_netifInit,
+	.state_sz = sizeof(greth_state_t),
+	.state_align = _Alignof(greth_state_t),
+	.name = "greth",
+	.media = NULL
```
julianuziemblo, line 555: you don't have to set media explicitly - it's already default-initialized to 0 as it's a `static` variable

drivers/netif-driver.c
```diff
@@ -75,7 +75,7 @@ static err_t netif_dev_init(struct netif *netif)
 
 	LWIP_ASSERT("netif != NULL", (netif != NULL));
 
-	if (!strcmp(storage->drv->name, "enet") || !strcmp(storage->drv->name, "rtl")) {
+	if (!strcmp(storage->drv->name, "enet") || !strcmp(storage->drv->name, "rtl") || !strcmp(storage->drv->name, "greth")) {
```
julianuziemblo, line 78: I'm hesitant as this isn't your code, but this is MISRA violation (should be: `strcmp(...) != 0` instead of `!strcmp(...)`), so I'd say it should be refactored (in separate commit)

drivers/greth.c
```diff
@@ -0,0 +1,562 @@
...
+#include <stdarg.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include "greth-regs.h"
+#include <sys/interrupt.h>
+
+#define greth_printf(state, fmt, ...) printf("lwip: greth@%08x: " fmt "\n", state->dev_phys_addr, ##__VA_ARGS__);
```
julianuziemblo, line 37: nitpick: semicolon at the end not needed

drivers/ephy.h
```diff
@@ -40,6 +46,8 @@ typedef struct {
 int ephy_init(eth_phy_state_t *phy, char *conf, uint8_t board_rev, link_state_cb_t cb, void *cb_arg);
 int ephy_linkSpeed(const eth_phy_state_t *phy, int *full_duplex);
 
+void ephy_macInterrupt(eth_phy_state_t *phy);
```
julianuziemblo, line 49: `*phy` should be `const` as it's not changed by the function

drivers/ephy.c
```diff
@@ -349,32 +390,63 @@ static void ephy_restartAN(const eth_phy_state_t *phy)
 		/* don't adv: 1000Base-T EEE (MMD write) */
 		ephy_mmdWrite(phy, 0x7, 0x3c /* EEEAR */, 0);
 	}
+	if (phy->model == ephy_ksz9031mnx) {
+		/* adv: 1000M-FD */
+		ephy_regWrite(phy, EPHY_KSZ9031_09_GBCR, (1u << 9));
```
julianuziemblo, line 395: I think registers 0x0 - 0xf are generic across PHYs, so this one should be transferred to the `COMMON` section - I've added it to the specific section as I didn't know it would be used by a different PHY

julianuziemblo (changes requested):
I like the GPIO handling (for now), I'd love to make it cleaner, but I don't think we can if we don't want to modularize the irq and reset handling in PHY (which would be an overkill for now in my opinion).

drivers/ephy.c
```diff
@@ -521,8 +595,10 @@ static int ephy_config(eth_phy_state_t *phy, char *cfg)
 	}
 
 	if (*p == '\0') {
+		phy->irq = EPHY_MAC_INTERRUPT;
 		return 0;
 	}
+	phy->irq = EPHY_GPIO_INTERRUPT;
```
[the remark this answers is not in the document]
julianuziemblo: also also: I'm not sure this variable is even needed as it's only checked once in `ephy_init`, and we can just check the `phy->model` instead - @lukileczo what do you think?

drivers/ephy.h
julianuziemblo: I'd also add a comment explaining the purpose and usage of this function

drivers/ephy.c
```diff
@@ -352,32 +392,63 @@ static void ephy_restartAN(const eth_phy_state_t *phy)
...
+		case ephy_rtl8201fi:
+			irq_reg = EPHY_RTL8201_1E_IISDR;
+			break;
+		case ephy_rtl8211fdi:
+			irq_reg = EPHY_RTL8211_1D_INSR;
+			break;
+		default:
+			/* unreachable */
+			return;
+	}
```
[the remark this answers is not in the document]
julianuziemblo: Maybe make it like this (to not repeat code):

```c
struct {
    enum { ... } type;
    uint16_t reg;
    uint16_t mask;
} irq;
```

drivers/ephy.c
```diff
@@ -770,36 +847,56 @@ int ephy_init(eth_phy_state_t *phy, char *conf, uint8_t board_rev, link_state_cb
...
+			break;
+		case ephy_rtl8211fdi:
+			phy->irq_mask = 0x06BD;
+			break;
+		default:
+			/* unreachable */
+			break;
+	}
+
+	if (phy->irq != ephy_mac_irq) {
```
julianuziemblo, line 868: I think more semantically correct thing is to check if irq is `gpio_interrupt`

drivers/greth.c
```diff
@@ -0,0 +1,562 @@
...
+			s->mmio->CTRL = ctr;
+		}
+		if (speed == 1000) {
+			/* Half duplex not supported in Gbit mode */
+			uint32_t ctr = s->mmio->CTRL;
+			ctr |= GRETH_CTRL_FD;
+			ctr &= ~(GRETH_CTRL_SP);
+			ctr |= GRETH_CTRL_GB;
+			s->mmio->CTRL = ctr;
+		}
```
julianuziemblo, line 229:
```suggestion
                uint32_t ctrl = s->mmio->CTRL & ~(GRETH_CTRL_FD | GRETH_CTRL_SP | GRETH_CTRL_GB);
                if (full_duplex != 0) {
                    ctrl |= GRETH_CTRL_FD;
                }
                switch (speed) {
                    case 100:
                        ctrl |= GRETH_CTRL_SP;
                        break;
                    case 1000:
                        /* Half duplex not supported in Gbit mode */
                        ctrl |= GRETH_CTRL_GB | GRETH_CTRL_FD;
                        break;
                    default:
                        /* do nothing */
                        break;
                }
                s->mmio->CTRL = ctrl;

File: drivers/greth.c [hunk 29]
```diff
@@ -0,0 +1,566 @@
...
+	}
+
+	return 1;
+}
+
+
+static void greth_irqThread(void *arg)
+{
+	greth_state_t *state = arg;
+	size_t rx_done = 0;
```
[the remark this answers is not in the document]
atlomak: If so, should it also be modified it the `imx-enet.c`?

drivers/greth.c
atlomak: done

drivers/greth.c
atlomak: done

drivers/greth.c
atlomak: done

drivers/greth.c
atlomak: done, changed to boolean

drivers/greth.c
atlomak: done, added comment and changed it to compile-time check

drivers/greth.c
```diff
@@ -0,0 +1,566 @@
...
+static int greth_irqHandler(unsigned int n, void *arg)
+{
+	(void)n;
+	greth_state_t *state = arg;
+
+	if ((state->mmio->STAT & (GRETH_STAT_TA | GRETH_STAT_RA)) != 0) {
+		state->drv_exit = true;
+	}
+
+	return 1;
```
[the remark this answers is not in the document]
atlomak: done

drivers/greth.c
```diff
@@ -0,0 +1,562 @@
...
+}
+
+
+static int greth_createResources(greth_state_t *state)
+{
+	if (mutexCreate(&state->irq_lock) < 0) {
+		return -1;
+	}
+	if (condCreate(&state->irq_cond) < 0) {
+		return -1;
```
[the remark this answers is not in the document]
atlomak: added initialization, now mutex is destroyed in `greth_cleanupResources`

drivers/greth.c
```diff
@@ -0,0 +1,564 @@
...
+		};
+		net_bufdesc_ring_t rings[2];
+	};
+
+	eth_phy_state_t phy;
+
+} greth_state_t;
+
+
+static int greth_disable_edcl(greth_state_t *state)
```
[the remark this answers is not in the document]
atlomak: fixed

drivers/ephy.c
```diff
@@ -272,6 +282,34 @@ static inline int ephy_ksz8081rnx_linkSpeed(const eth_phy_state_t *phy, int *ful
...
+	}
+	else if ((pc & (1u << 5)) != 0) {
+		return 100;
+	}
+	else if ((pc & (1u << 6)) != 0) {
+		return 1000;
+	}
+	else {
+		return 0;
+	}
```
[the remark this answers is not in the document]
julianuziemblo: That's valid. The function doesn't seem to be a bottleneck, so this should be fine also

drivers/greth.c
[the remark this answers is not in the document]
julianuziemblo: yes

drivers/greth.c
julianuziemblo: good idea, so future maintainers don't take it as an error

drivers/greth.c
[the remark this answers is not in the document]
julianuziemblo: I don't think it is: it was created successfully, so `state->irq_lock` has a value, and `greth_cleanupResources` checks the value before destroying it, so it will not be destroyed. `state->irq_lock` should be set explicitly to -1 if `condCreate` fails, or destroyed before returning -1 from `greth_resourceCreate`

drivers/greth.c
```diff
@@ -0,0 +1,562 @@
...
+	if (state->irq_lock < (handle_t)-1) {
+		resourceDestroy(state->irq_lock);
+	}
+	if (state->irq_cond < (handle_t)-1) {
+		resourceDestroy(state->irq_cond);
+	}
+}
+
+
+static int greth_initDevice(greth_state_t *state, int irq, int mdio)
```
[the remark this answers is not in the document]
julianuziemblo: Some lines down, `mdio` is still checked as `mdio != 0` - that should be fixed too.

Also: remember to change it in other files too (`drivers/imx-enet.c` and `drivers/rtl8139cp.c`)!

drivers/greth.c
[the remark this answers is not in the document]
julianuziemblo: okay, I was looking at the wrong code, this is fine

drivers/greth-regs.h
```diff
@@ -0,0 +1,120 @@
...
+#define GRETH_RX_DESC_BASEADDR_SHIFT (10)
+#define GRETH_RX_DESC_BASEADDR_MASK  BITS(GRETH_RX_DESC_BASEADDR_SHIFT, 31) /* TX descriptor base address */
+#define GRETH_RX_DESC_DESCPNT_SHIFT  (3)
+#define GRETH_RX_DESC_DESCPNT_MASK   BITS(GRETH_RX_DESC_DESCPNT_SHIFT, 9) /* TX descriptor pointer */
+	uint32_t EDCL_IP;                                                     /* 0x1C EDCL IP */
+	uint32_t HASH_MSB;                                                    /* 0x20 Hash table MSB */
+	uint32_t HASH_LSB;                                                    /* 0x24 Hash table LSB */
+	uint32_t EDCL_MAC_MSB;                                                /* 0x28 EDCL MAC address MSB */
+	uint32_t EDCL_MAC_LSB;                                                /* 0x2C EDCL MAC address LSB */
+	R_RESERVED(0x30, 0xff);                                               /* 0x30 - 0xFF Reserved */
```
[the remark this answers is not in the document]
atlomak: I'd rather keep it, as it clearly resembles registers layout from the docs.

drivers/greth.c
```diff
@@ -0,0 +1,601 @@
...
+	}
+	if (state->irq_cond != (handle_t)-1) {
+		resourceDestroy(state->irq_cond);
+	}
+}
+
+
+static int greth_initDevice(greth_state_t *state, int irq, bool mdio)
+{
+	state->mmio = physmmap(state->dev_phys_addr, 0x1000);
```
julianuziemblo, line 427: What is the page size on gr740? 0x1000 seems like an overkill

drivers/greth.c
```diff
@@ -0,0 +1,566 @@
...
+		state->mmio->STAT = GRETH_STAT_TI;
+		net_reapTxFinished(&state->tx);
+
+		if ((state->mmio->STAT & GRETH_STAT_PS) != 0) {
+			state->mmio->STAT = GRETH_STAT_PS;
+			ephy_macInterrupt(&state->phy);
+		}
+
+		if ((state->mmio->STAT & (GRETH_STAT_TI | GRETH_STAT_RI | GRETH_STAT_PS)) == 0) {
+			state->mmio->CTRL |= GRETH_CTRL_TI | GRETH_CTRL_RI | GRETH_CTRL_PI;
```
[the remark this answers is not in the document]
atlomak: I've added mask lifting. Also added checks for `tx` or `rx` interrupts to avoid mutex acquisition and save some cycles.

drivers/greth-regs.h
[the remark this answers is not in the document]
atlomak: ok, removed

drivers/greth.c
```diff
@@ -0,0 +1,624 @@
...
+	state->mmio = physmmap(state->dev_phys_addr, _PAGE_SIZE);
+	if (state->mmio == MAP_FAILED) {
+		return -ENOMEM;
+	};
+
+	greth_reset(state);
+
+#if GRETH_DISABLE_EDCL
+	greth_disableEdcl(state);
+#endif
```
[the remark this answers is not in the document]
julianuziemblo: why wrap it in `(   != 0)`, if it's already compatible with MISRA? 

If you worry about the "user" setting it not-MISRA-lke to e.g. 5 in `board_config.h`, maybe check if `GRETH_DISABLE_EDCL == 0 || GRETH_DISABLE_EDCL == 1`? 

To me, `(GRESTH_DISABLE_EDCL != 0)` suggests that it can have other values than 0 and 1, despite its name ("enable"/"disable" -> boolean)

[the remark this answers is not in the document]
lukileczo: Ok, maybe it's even better to just change it to `#ifdef` instead

drivers/greth.c
```diff
@@ -0,0 +1,624 @@
...
+
+
+static void greth_setCardMac(greth_state_t *state)
+{
+	uint64_t mac = GRETH_MAC_IAB;
+#ifdef GRETH_EUI48
+	mac |= (GRETH_EUI48 & 0xFFF);
+#endif
+	state->mmio->MAC_MSB = mac >> 32;
+	state->mmio->MAC_LSB = mac & 0xFFFFFFFF;
```
lukileczo, line 369:
```suggestion
	state->mmio->MAC_LSB = mac & 0xFFFFFFFFu;
```

drivers/greth.c
```diff
@@ -0,0 +1,624 @@
...
+{
+	return (v >> (i * 8)) & 0xFF;
+}
+
+
+static void greth_setCardMac(greth_state_t *state)
+{
+	uint64_t mac = GRETH_MAC_IAB;
+#ifdef GRETH_EUI48
+	mac |= (GRETH_EUI48 & 0xFFF);
```
lukileczo, line 366:
```suggestion
	mac |= (GRETH_EUI48 & 0xFFFu);
```

lukileczo (approved):
LGTM

drivers/greth.c
```diff
@@ -0,0 +1,631 @@
...
+		resourceDestroy(state->irq_cond);
+	}
+}
+
+
+static void greth_reset(greth_state_t *state)
+{
+	state->mmio->CTRL = GRETH_CTRL_RS;
+	do {
+	} while ((state->mmio->CTRL & GRETH_CTRL_RS) != 0);
```
julianuziemblo, line 447: just `while (...)`

drivers/greth.c
```diff
@@ -0,0 +1,631 @@
...
+static int greth_initDevice(greth_state_t *state, int irq, bool mdio)
+{
+	state->mmio = physmmap(state->dev_phys_addr, _PAGE_SIZE);
+	if (state->mmio == MAP_FAILED) {
+		return -ENOMEM;
+	};
+
+	greth_reset(state);
+
+#if (GRETH_EDCL != 1)
```
julianuziemblo, line 460: MISRA: `GRETH_EDCL` is a macro, so it can just be checked with `#if !GRETH_EDCL`

drivers/greth.c
```diff
@@ -0,0 +1,631 @@
...
+	state = netif->state;
+	state->netif = netif;
+
+	if (cfg == NULL) {
+		return -EINVAL;
+	}
+
+	/* base addr */
+	state->dev_phys_addr = strtoul(cfg, &p, 0);
+	if (*cfg == '\0' || *p++ != ':') {
```
julianuziemblo, line 531: MISRA: statements should not have side effects, so `p++` should be after the `if` as it runs after the condition is checked

drivers/greth.c
```diff
@@ -0,0 +1,631 @@
...
+		return -EINVAL;
+	}
+
+	/* irq */
+	cfg = p;
+	irq = strtoul(cfg, &p, 0);
+	if (*cfg == '\0' || (*p != '\0' && *p++ != ':') || irq < 0) {
```
julianuziemblo, line 538: MISRA: statements should not have side effects (`p++` should be after if)

drivers/greth.c
```diff
@@ -0,0 +1,631 @@
...
+		return -EINVAL;
+	}
+
+	/* MDIO and PHY opts */
+	cfg = NULL;
+	while (p != NULL && *p != '\0') {
+		cfg = strchr(p, ':');
+		if (cfg != NULL) {
+			*cfg++ = '\0';
```
julianuziemblo, line 547: MISRA: statements should not have side effects
```suggestion
			*cfg = '\0';
			cfg++;
```

drivers/greth.c
```diff
@@ -0,0 +1,631 @@
...
+	if (err != 0) {
+		return err;
+	}
+	greth_debug_printf(state, "Initialized GRETH");
+
+	if (cfg != NULL) {
+		err = ephy_init(&state->phy, cfg, 0, greth_setLinkState, (void *)state->netif);
+		if (err < 0) {
+			greth_printf(state, "WARN: PHY init failed: %s (%d)", strerror(-err), err);
+			return err;
```
julianuziemblo, line 573:
```suggestion
            greth_resourceCleanup(state);
			return err;
```

julianuziemblo (changes requested):
Some more MISRA and a few nitpicks, beside those - looks good

README.md
```diff
@@ -25,14 +25,16 @@ For qemu, it's usually: `rtl:0x10:10`
 
 When an external PHY is used, *phyinfo* is formatted as follows:
 
-	PHY[:model]:[busnr.]id[:reset:[-]n:/dev/gpioX][:irq:[-]n:/dev/gpioX]
+	PHY[:model]:[busnr.]id[:irq:MAC][[:reset:[-]n:/dev/gpioX][:irq:[-]n:/dev/gpioX]]
```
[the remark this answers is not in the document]
julianuziemblo: still, it would be clearer to define only one section of IRQ as opposed to 2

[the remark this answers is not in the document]
julianuziemblo: I'd see it all together like so:

`PHY[:model]:[busnr.]id[:irq:{MAC|[-]n:/dev/gpioX}][:reset:[-]n:/dev/gpioX]`

Because I could see a situation where only the reset pin is wired and irq pin is MAC, as these are two separate parts. If you don't think we should support that, I'd at least merge the IRQ options.

Overall, this whole string belongs in `board_config.h`, but that's a topic for another pull request :smile:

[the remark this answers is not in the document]
atlomak: I have merged both IRQ options. For now, without a support for both reset and MAC, since there isn't board that uses it yet.

julianuziemblo (approved):
LGTM

PR #118, opened 2025-10-15
wi-fi: Add CYW4373E USB WiFi modem support

julianuziemblo:
`clang-format-diff` and `codespell` checked locally on this PR's latest commit to be passing as CI fails for these jobs because of its' size.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `armv7m7-imxrt117x-evk`, `armv7a7-imx6ull-dataro`

## Checklist:

## Special treatment

    - https://github.com/phoenix-rtos/phoenix-rtos-devices/pull/596


nalajcie:
was the new version of WHD driver tested on DTR platform?

wi-fi/whd/whd_chip.c
```diff
@@ -88,1252 +103,1719 @@ static whd_result_t whd_enable_save_restore(whd_driver_t whd_driver);
...
+        {
+            /* refer to whd_cdc_get_iovar_buffer() */
+            while (!*data)
+            {
+                whd_driver->whd_ioctl_log[i].data_size--;
+                data++;
+            }
+
+            if (strlen( (char *)data ) <= WHD_IOVAR_STRING_SIZE)
+                strcpy(iovar, (char *)data);
```
ziemleszcz: This is a potential buffer overflow that was fixed in my original port. See commit 024f82240e2b60e7d608c4e1316ae0c146221f6b.

wi-fi/hal/cyabs_rtos_impl.h
```diff
@@ -65,14 +66,14 @@ extern "C" {
...
-	CY_RTOS_PRIORITY_REALTIME = 6,    /**< Realtime Thread priority */
-	CY_RTOS_PRIORITY_MAX = 7          /**< Maximum allowable Thread priority */
+	CY_RTOS_PRIORITY_MIN = 7,         /**< Minimum allowable Thread priority */
+	CY_RTOS_PRIORITY_LOW = 6,         /**< A low priority Thread */
+	CY_RTOS_PRIORITY_BELOWNORMAL = 5, /**< A slightly below normal Thread priority */
+	CY_RTOS_PRIORITY_NORMAL = 4,      /**< The normal Thread priority */
+	CY_RTOS_PRIORITY_ABOVENORMAL = 3, /**< A slightly elevated Thread priority */
+	CY_RTOS_PRIORITY_HIGH = 2,        /**< A high priority Thread */
+	CY_RTOS_PRIORITY_REALTIME = 1,    /**< Realtime Thread priority */
+	CY_RTOS_PRIORITY_MAX = 0          /**< Maximum allowable Thread priority */
```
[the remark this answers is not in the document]
julianuziemblo: this is how it should have been as in Phoenix-RTOS priority `0` is the "highest" priority

wi-fi/hal/cyhal_usb.c
```diff
@@ -0,0 +1,344 @@
...
+			msg.o.size = buflen;
+			break;
+
+		case usbwlan_abort:
+			is_tx = false;
+			break;
+
+		default:
+			return -EOPNOTSUPP;
+	}
```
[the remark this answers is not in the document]
julianuziemblo: ... but there is a default case in the code you mention?

wi-fi/whd/bus_protocols/whd_bus_protocol_interface.h
```diff
@@ -75,44 +76,44 @@ typedef void (*whd_bus_irq_callback_t)(void *handler_arg, uint32_t event);
...
+                                           uint8_t value_length, uint8_t *value);
 extern whd_result_t whd_bus_mem_bytes(whd_driver_t whd_driver, uint8_t direct,
-	uint32_t address, uint32_t size, uint8_t *data);
+                                      uint32_t address, uint32_t size, uint8_t *data);
 extern whd_driver_t whd_bt_get_whd_driver(void);
 extern whd_result_t whd_bus_share_bt_init(whd_driver_t whd_driver);
 extern whd_result_t whd_bus_bt_attach(whd_driver_t whd_driver, void *btdata,
-	void (*bt_int_fun)(void *data));
+                                      void (*bt_int_fun)(void *data) );
 extern void whd_bus_bt_detach(whd_driver_t whd_driver);
```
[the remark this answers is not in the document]
julianuziemblo: ... thanks? I didn't change it though - it's from the new WHD version.

wi-fi/hal/cyhal_usb.c
```diff
@@ -0,0 +1,344 @@
...
+
+cy_rslt_t cyhal_usb_set_chip(const cyhal_usb_t *obj, uint32_t chip)
+{
+	if (obj == NULL || obj->usb_priv == NULL) {
+		return CY_RTOS_BAD_PARAM;
+	}
+	const usb_priv_t *priv = obj->usb_priv;
+	const usbwlan_i_t imsg = { .type = usbwlan_setChip };
+	int res = cyhal_usb_devctl(priv, &chip, sizeof(chip), &imsg);
+	printf("result is %d\n", res);
```
Darchiv: Leftover printf. Either remove or use `cy_log_msg()`

wi-fi/hal/cyabs_rtos.c
```diff
@@ -265,3 +266,105 @@ cy_rslt_t cy_rtos_delay_milliseconds(cy_time_t num_ms)
...
+		return CY_RTOS_NO_MEMORY;
+	}
+
+	queue->length = length;
+	queue->itemsz = itemsize;
+	atomic_init(&queue->head, 0);
+	atomic_init(&queue->tail, 0);
+
+	return CY_RSLT_SUCCESS;
+}
```
Darchiv, line 319: `cy_rtos_queue_*()` looks like a nice place for using our implementation of a ring buffer: https://github.com/phoenix-rtos/phoenix-rtos-corelibs/blob/master/libalgo/lf-fifo.h

I understand that not all use cases may be possible, as lf-fifo does not provide low-level access routines, so this is more of a future refinement. If this usage seems sensible, then you can poke @ziemleszcz to implement the missing functionality.

wi-fi/hal/cyhal_usb.c
```diff
@@ -0,0 +1,344 @@
...
+	switch (imsg->type) {
+		case usbwlan_ctrlOut:
+		case usbwlan_regWrite:
+			is_tx = true;
+			msg.i.data = buffer;
+			msg.i.size = buflen;
+			break;
+
+		case usbwlan_setChip:
+			is_tx = false;
```
Darchiv: This case also uses `msg.i`, the same as above, but `is_tx` is `false`. Why is that?

Nit: `tx` seems to be a bit off in relation to the surrounding code (i.e. you operate on message abstraction which uses `msg.i` and `msg.o` - input and output buffers) - maybe use something like `is_output` or `is_write`?

wi-fi/hal/cyhal_usb.c
```diff
@@ -0,0 +1,344 @@
...
+		obj->usb_priv = NULL;
+	}
+
+	return CY_RSLT_SUCCESS;
+}
+
+
+static inline cy_rslt_t cyhal_usb_convert_result(ssize_t result)
+{
+	return (result >= 0) ? CY_RSLT_SUCCESS : CYHAL_USB_RSLT_ERR_DEVICE_OP(result);
```
Darchiv: Is `CYHAL_USB_RSLT_ERR_DEVICE_OP()` prepared for negative integers? In other places you pass `errno` directly, which is a non-negative number

wi-fi/hal/cyhal_usb.h
```diff
@@ -0,0 +1,152 @@
...
+#define CYHAL_USB_DL_STATE_BAD_CRC      3 /* compressed image was corrupted */
+#define CYHAL_USB_DL_STATE_RUNNABLE     4 /* download was successful,waiting for go cmd */
+#define CYHAL_USB_DL_STATE_START_FAIL   5 /* failed to initialize correctly */
+#define CYHAL_USB_DL_STATE_NVRAM_TOOBIG 6 /* host specified nvram data exceeds DL_NVRAM value */
+#define CYHAL_USB_DL_STATE_IMAGE_TOOBIG 7 /* firmware image too big */
+
+#define WHD_USB_MAX_BULK_TRANSFER_SIZE (512) /* Max packet size for high speed USB device */
+
+#define CYHAL_USB_RSLT_ERR_DEVICE_NOT_FOUND CY_RSLT_CREATE(CY_RSLT_TYPE_ERROR, CYHAL_RSLT_MODULE_USB, 0x1)
+#define CYHAL_USB_RSLT_ERR_DEVICE_OP(err)   CY_RSLT_CREATE(CY_RSLT_TYPE_ERROR, CYHAL_RSLT_MODULE_USB, err)
```
Darchiv, line 57: Is `err` supposed to be errno-compatible? Because it sometimes is used as such and on the line above you use `0x1` as a custom value for `CYHAL_USB_RSLT_ERR_DEVICE_NOT_FOUND`, which maps to `errno = EPERM` - unless you differentiate between negative and positive error codes, but then please put a comment that this is the intended behavior and ensure that signed numbers are correctly put into **and** correctly interpreted by other parts of WHD. (See my other comment on that.)

wi-fi/hal/cyhal_usb.c
```diff
@@ -0,0 +1,344 @@
...
+{
+	int result;
+	__attribute__((unused)) oid_t oid;
+
+	do {
+		result = lookup(path, NULL, &oid);
+		usleep(10 * 1000);
+	} while ((result >= 0) != state);
+
+	return CY_RSLT_SUCCESS;
```
Darchiv, line 110: Is this function expected to block indefinitely? Does reinitialization (e.g. on USB bus reset, re-enumeration, user abort) work correctly and does not hang?

julianuziemblo: 1) yes
2) AFAIK yes - tried test multiple corner cases

wi-fi/hal/cyhal_usb.h
julianuziemblo: I decided it should be errno-compatible. Made `...._DEVICE_NOT_FOUND` take in `ENODEV` as parameter and `..._DEVICE_OP(err)` be initialized with `errno` or `abs(retval)` on failure. What do you think?

wi-fi/hal/cyabs_rtos.c
julianuziemblo: used it, works fine

wi-fi/whd/whd_chip.c
julianuziemblo: I tried to fix those issues that were there in separate commits, but I didn't look into the original "port" commit. Are there any more of those you fixed there that should also be fixed here?

wi-fi/whd/whd_debug.h
```diff
@@ -39,25 +40,19 @@ extern "C" {
...
 #else
-#define WHD_TRIGGER_BREAKPOINT()
+#define WHD_TRIGGER_BREAKPOINT( )
 #endif
 
 #ifdef WPRINT_ENABLE_ERROR
 #define WPRINT_ERROR(args) \
 	do { \
-		WPRINT_MACRO(args); \
+		WPRINT_ERROR(args); \
```
ziemleszcz: ???

wi-fi/whd/whd_thread.c
```diff
@@ -268,77 +364,91 @@ void whd_thread_notify(whd_driver_t whd_driver)
...
+        if (rx_cnt >= WHD_THREAD_RX_BOUND)
+        {
+            thread_info->bus_interrupt = WHD_TRUE;
+            rx_over_bound = 1;
+            continue;
+        }
+        if (bus_fail > WHD_MAX_BUS_FAIL)
+        {
+            WPRINT_WHD_ERROR( ("%s: Error bus_fail over %d times, exiting\n", __FUNCTION__, WHD_MAX_BUS_FAIL) );
+#if 0
```
ziemleszcz: Why `#if 0` was added?

wi-fi/whd/whd_chip.c
ziemleszcz: I didn't found anything else.

wi-fi/whd/whd_debug.h
julianuziemblo: hm, thought I just re-applied your changes. Probably missed this, thanks

wi-fi/whd/whd_thread.c
julianuziemblo: it doesn't have to be, I just wanted to show that we're either closing or reporting events. Do you think this is ok, or should I e.g. delete this entirely or maybe turn it back on?

wi-fi/whd/whd_debug.h
ziemleszcz: I don't think it should be removed. `WPRINT_ERROR` macro is used by other macros below. The problem was that is `WPRINT_ERROR` was recursive.

wi-fi/whd/whd_thread.c
julianuziemblo: deleted it

wi-fi/hal/cyabs_rtos_impl.h
```diff
@@ -28,9 +28,12 @@
 
 #pragma once
 
+#include <stdatomic.h>
 #include <sys/types.h>
 #include <sys/threads.h>
 
+#include "lf-fifo.h"
```
ziemleszcz: <lf-fifo.h>

wi-fi/hal/cyabs_rtos.c
```diff
@@ -17,6 +17,9 @@
 #include <time.h>
 #include <unistd.h>
 #include <stdlib.h>
+#include <string.h>
+
+#include "lf-fifo.h"
```
ziemleszcz: <lf-fifo.h>

wi-fi/hal/cyabs_rtos.c
```diff
@@ -265,3 +268,74 @@ cy_rslt_t cy_rtos_delay_milliseconds(cy_time_t num_ms)
 
 	return CY_RSLT_SUCCESS;
 }
+
+
+static inline bool cy_rtos_queue_is_power_of_2(size_t n)
+{
+	return (n & (n - 1)) == 0;
```
ziemleszcz: Nitpick: returns true for 0.

wi-fi/hal/cyabs_rtos.c
```diff
@@ -265,3 +268,74 @@ cy_rslt_t cy_rtos_delay_milliseconds(cy_time_t num_ms)
...
+}
+
+
+/* preconditions:
+ * - length must be a power of 2
```
ziemleszcz: and `>= 2`

wi-fi/hal/cyabs_rtos.c
```diff
@@ -265,3 +268,74 @@ cy_rslt_t cy_rtos_delay_milliseconds(cy_time_t num_ms)
...
+ * - if queue->data != NULL, it should point to an array of at least length * itemsize bytes
+ */
+cy_rslt_t cy_rtos_queue_init(cy_queue_t *queue, size_t length, size_t itemsize)
+{
+	if (queue == NULL || !cy_rtos_queue_is_power_of_2(length) || length == 0 || itemsize == 0) {
```
ziemleszcz: `length >= 2`

wi-fi/hal/cyabs_rtos.c
```diff
@@ -265,3 +268,74 @@ cy_rslt_t cy_rtos_delay_milliseconds(cy_time_t num_ms)
...
+	lf_fifo_init(&queue->fifo, queue->fifo.data, size);
+	queue->itemsz = itemsize;
+
+	return CY_RSLT_SUCCESS;
+}
+
+
+cy_rslt_t cy_rtos_queue_put(cy_queue_t *queue, const void *item_ptr)
+{
+	return lf_fifo_push_many(&queue->fifo, item_ptr, queue->itemsz) == 0 ? CY_RTOS_QUEUE_FULL : CY_RSLT_SUCCESS;
```
ziemleszcz, line 327: Technically, this operation may not be atomic. If `itemsz > 1`, `lf_fifo_push_many()` may push only part of the item.
I understand this is mitigated by configuring FIFO size as a multiple of `itemsz`, but it would be helpful to add a comment explaining why this works.
Additionally, you may still consider using `lf_fifo_free()` to verify that there is sufficient space to push the complete item.

wi-fi/whd/bus_protocols/whd_bus_m2m_protocol.c
```diff
@@ -0,0 +1,789 @@
...
+    while ( ( (whd_bus_m2m_read_backplane_value(whd_driver, M2M_REG_DAR_D2H_MSG_0, 1, &byte_data) ) == 0 ) &&
+            ( (byte_data & state) == 0 )  &&
+            (loop_count < 30000) )
+    {
+        loop_count += 10;
+    }
+    if (loop_count >= 30000)
+    {
+        WPRINT_WHD_ERROR( ("%s: D2H Wait TimeOut! \n", __FUNCTION__) );
+        return -1;
```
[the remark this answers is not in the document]
julianuziemblo: true, but M2M is not a part of this PR and probably will not even be used

wi-fi/hal/cyabs_rtos.c
```diff
@@ -265,3 +268,78 @@ cy_rslt_t cy_rtos_delay_milliseconds(cy_time_t num_ms)
 
 	return CY_RSLT_SUCCESS;
 }
+
+
+static inline bool cy_rtos_queue_is_power_of_2(size_t n)
+{
+	return (n & (n - 1)) == 0;
+}
```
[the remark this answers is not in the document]
julianuziemblo: good catch, but 1) it's not used for 0 anyway, and 2) the "nice trick" is already there - you just added the extra 0 check ;)

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -216,37 +283,158 @@ whd_result_t whd_bus_usb_init(whd_driver_t whd_driver)
  **************************************************************************************************/
 whd_result_t whd_bus_usb_deinit(whd_driver_t whd_driver)
 {
+    if (whd_bus_is_up(whd_driver) == WHD_FALSE) {
+        return WHD_SUCCESS;
+    }
+    whd_bus_set_state(whd_driver, WHD_FALSE);
+    cy_rslt_t cy_result = cyhal_usb_free(whd_driver->bus_priv->usb_obj);
+    whd_result_t whd_result = whd_bus_usb_stop_rx(whd_driver);
```
ziemleszcz, line 291: Shouldn't RX thread be terminated before freeing `usb_obj`? Otherwise, it may still access it, leading to a use-after-free issue.

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -505,14 +682,76 @@ whd_result_t whd_bus_usb_read_frame(whd_driver_t whd_driver, whd_buffer_t* buffe
...
+                (unsigned short)(WHD_USB_MAX_RECEIVE_BUF_SIZE + (uint16_t)sizeof(whd_buffer_header_t)),
+                (whd_sdpcm_has_tx_packet(whd_driver) ? 0 : WHD_RX_BUF_TIMEOUT)) != WHD_SUCCESS) {
+        return WHD_BUFFER_ALLOC_FAIL;
+    }
+
+    void *data = whd_buffer_get_current_piece_data_pointer(whd_driver, *buffer);
+    size_t buflen = WHD_USB_MAX_RECEIVE_BUF_SIZE;
+    cy_rslt_t result = cyhal_usb_ctrl_receive(whd_driver->bus_priv->usb_obj, data, &buflen);
+
+    return (result == CY_RSLT_SUCCESS) ? WHD_SUCCESS : WHD_HAL_ERROR;
```
ziemleszcz: Buffer leak: on `WHD_HAL_ERROR`, the allocated buffer is not freed by the caller (it returns immediately via `CHECK_RETURN`).

wi-fi/hal/cyhal_usb.c
```diff
@@ -0,0 +1,335 @@
...
+			is_tx = false;
+			break;
+
+		default:
+			return -EOPNOTSUPP;
+	}
+
+	memcpy(msg.i.raw, imsg, sizeof(*imsg));
+
+	int length = msgSend(priv->oid.port, &msg);
```
ziemleszcz: Misleading, it's not `length`. Maybe rename it to `res/ret`.

wi-fi/hal/cyhal_usb.c
```diff
@@ -0,0 +1,335 @@
...
+		usleep(10 * 1000);
+	} while (check != state);
+
+	return CY_RSLT_SUCCESS;
+}
+
+
+static ssize_t cyhal_usb_devctl(const usb_priv_t *priv, void *buffer, size_t buflen, const usbwlan_i_t *imsg)
+{
+	bool is_tx;
```
ziemleszcz: Set to `false` by default?

wi-fi/hal/cyhal_usb.c
```diff
@@ -0,0 +1,335 @@
...
+		case usbwlan_ctrlIn:
+		case usbwlan_regRead:
+		case usbwlan_dl:
+			is_tx = false;
+			msg.o.data = buffer;
+			msg.o.size = buflen;
+			break;
+
+		case usbwlan_abort:
+			is_tx = false;
```
ziemleszcz, line 139: What is the purpose of this ioctl?

wi-fi/hal/cyhal_usb.c
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	const usb_priv_t *priv = obj->usb_priv;
+	const usbwlan_i_t umsg = { .type = usbwlan_ctrlOut };
+	ssize_t status;
+	cy_rslt_t result;
+
+	for (;;) {
+		status = cyhal_usb_devctl(priv, buffer, *buflen, &umsg);
+		if (status < 0 && (errno == EAGAIN || errno == EINTR)) {
```
ziemleszcz: `cyhal_usb_devctl` doesn't set errno.

wi-fi/hal/cyhal_usb.c
```diff
@@ -0,0 +1,335 @@
...
+		return CY_RTOS_BAD_PARAM;
+	}
+
+	const usb_priv_t *priv = obj->usb_priv;
+	const usbwlan_i_t umsg = { .type = usbwlan_ctrlIn };
+	ssize_t status;
+	cy_rslt_t result;
+
+	for (;;) {
+		status = cyhal_usb_devctl(priv, buffer, *buflen, &umsg);
```
ziemleszcz, line 319: `cyhal_usb_devctl` doesn't set errno.

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -216,37 +283,158 @@ whd_result_t whd_bus_usb_init(whd_driver_t whd_driver)
...
+}
+
+
+static whd_result_t whd_bus_usb_stop_rx(whd_driver_t whd_driver)
+{
+    cy_rslt_t result = CY_RSLT_SUCCESS;
+    whd_bus_set_state(whd_driver, WHD_FALSE);
+    whd_usb_rx_thread_notify();
+    whd_thread_notify(whd_driver);
+    result |= cy_rtos_join_thread(&whd_bus_usb_common.rx.thread);
```
ziemleszcz, line 423: RX thread is not present on this path:
whd_bus_usb_init() -> whd_bus_usb_deinit() -> whd_bus_usb_stop_rx()

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -356,10 +543,8 @@ static whd_result_t whd_bus_usb_download_firmware(whd_driver_t whd_driver)
                                            /* size   */ &size,
                                            /* buffer */ buff));
 
-            /* Simply avoid having to send a ZLP by ensuring we never have an even
-             * multiple of 64 */
-            if (!(size % 64))
-            {
+            /* Simply avoid having to send a ZLP by ensuring we never have an even multiple of 64 */
+            if (size % 64 == 0) {
```
ziemleszcz, line 572: && size >= 4

wi-fi/hal/cyabs_rtos.c
```diff
@@ -265,3 +268,78 @@ cy_rslt_t cy_rtos_delay_milliseconds(cy_time_t num_ms)
...
 }
+
+
+static inline bool cy_rtos_queue_is_power_of_2(size_t n)
+{
+	return (n != 0) && (n & (n - 1)) == 0;
+}
+
+
+/* length * itemsz must be a power of 2 and >= 2
```
ziemleszcz: Move this comment to the header file?

wi-fi/hal/cyhal_usb.c
```diff
@@ -0,0 +1,340 @@
...
+
+cy_rslt_t cyhal_usb_free(cyhal_usb_t *obj)
+{
+	if (obj == NULL || obj->usb_priv == NULL) {
+		return CY_RTOS_BAD_PARAM;
+	}
+
+	usb_priv_t *priv = obj->usb_priv;
+
+	if (priv != NULL) {
```
ziemleszcz: Redundant check.

wi-fi/whd/whd_thread.c
```diff
@@ -222,28 +222,27 @@ int8_t whd_thread_receive_one_packet(whd_driver_t whd_driver)
...
-			{
-				whd_process_bdc_event(whd_driver, recv_buffer,
-										whd_buffer_get_current_piece_size(whd_driver, recv_buffer));
-			}
-			else
-			{
-				whd_process_bdc(whd_driver, recv_buffer);
-			}
+        if (data[0] == 0x20) {
+            /* Check if receive bdc Event or Data pack */
```
ziemleszcz, line 234: Space indentation.

wi-fi/hal/cyhal_usb.c
julianuziemblo: to forcibly abort the in-progress read, as closing the device while read is in progress does not work (close fails as file refcount in kernel is not 0)

wi-fi/whd/whd_wifi_api.h
```diff
@@ -365,6 +365,15 @@ extern whd_result_t whd_wifi_external_auth_request(whd_interface_t ifp,
  */
 extern whd_result_t whd_wifi_stop_external_auth_request(whd_interface_t ifp);
 
+/** Querry the chip ID with an IOCTL
```
ziemleszcz: Query

wi-fi/whd/whd_resources.c
```diff
@@ -268,9 +290,46 @@ uint32_t host_get_resource_block(whd_driver_t whd_drv, whd_resource_type_t type,
...
+	result = resource_lookup(whd_drv, type, &rsrc, &is_memory, &resource_size);
+	if (result != WHD_SUCCESS) {
+		return result;
+	}
+
+	if (offset >= resource_size) {
+		return WHD_BADARG;
+	}
+
+	if (offset + size >= resource_size) {
```
ziemleszcz: offset + size > resource_size

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
julianuziemblo: this shouldn't happen as the inner working of whd_resource_read() ensure we don't get chunks of size 0 which is the only way this can potentially fail. I'll add size == 0 check and break out of the loop in that case.

wi-fi/whd/whd_thread.c
julianuziemblo: rest of the file is space-indented, so I'll keep the space indentation

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,909 @@
...
+            cdc_header_t* cdc_header = (cdc_header_t*)(data + sdpcm_header_size);
+            uint32_t cdc_size = size - sdpcm_header_size;
+
+            /* Send control request */
+            CHECK_RETURN(whd_bus_usb_send_ctrl(whd_driver, cdc_header, &cdc_size));
+
+            /* Receive control response */
+            status = whd_bus_usb_receive_ctrl_buffer(whd_driver, &rec_buffer);
+            if (status == WHD_BUFFER_ALLOC_FAIL) {
+                whd_buffer_release(whd_driver, rec_buffer, WHD_NETWORK_RX);
```
ziemleszcz: Why are you freeing a non-existent buffer? If the intention was to free the buffer on `WHD_HAL_ERROR`, it would be better to handle this in `whd_bus_usb_receive_ctrl_buffer()` and keep the API unchanged.

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,909 @@
...
+    }
+
+    cy_rtos_exit_thread();
+}
+
+
+static whd_result_t whd_bus_usb_start_rx(whd_driver_t whd_driver)
+{
+    if (whd_bus_usb_common.rx.thread != 0) {
+        WPRINT_WHD_INFO(("USB Bus: RX thread already running\n"));
```
ziemleszcz, line 374: This code path may still result in a semaphore and queue leak.

wi-fi/whd/whd_thread.c
```diff
@@ -160,23 +213,63 @@ int8_t whd_thread_send_one_packet(whd_driver_t whd_driver)
...
+#if (CYBSP_WIFI_INTERFACE_TYPE == CYBSP_USB_INTERFACE)
+        uint8_t *data = whd_buffer_get_current_piece_data_pointer(whd_driver, recv_buffer);
+
+        if (data[0] == 0x20) {
+            /* Check if receive bdc Event or Data pack */
+            whd_event_t *event = (whd_event_t *)(data + sizeof(bdc_header_t));
+
+            if (ntoh16(event->eth.ethertype) == 0x886C /* <- ETHER_TYPE_BRCM */)
+            {
+            whd_process_bdc_event(whd_driver, recv_buffer,
```
ziemleszcz: Wrong indentation.

wi-fi/whd/whd_thread.c
```diff
@@ -160,23 +213,63 @@ int8_t whd_thread_send_one_packet(whd_driver_t whd_driver)
...
+        WHD_STATS_INCREMENT_VARIABLE(whd_driver, rx_total);
+
+#if (CYBSP_WIFI_INTERFACE_TYPE == CYBSP_USB_INTERFACE)
+        uint8_t *data = whd_buffer_get_current_piece_data_pointer(whd_driver, recv_buffer);
+
+        if (data[0] == 0x20) {
+            /* Check if receive bdc Event or Data pack */
+            whd_event_t *event = (whd_event_t *)(data + sizeof(bdc_header_t));
+
+            if (ntoh16(event->eth.ethertype) == 0x886C /* <- ETHER_TYPE_BRCM */)
```
ziemleszcz, line 237: This block mixes spaces and tabs for indentation. I’m not sure which style is correct here, but it should be consistent, please choose one.

wi-fi/hal/cyabs_rtos.h
```diff
@@ -469,6 +476,106 @@ cy_rslt_t cy_rtos_delay_milliseconds(cy_time_t num_ms);
...
+ * \ingroup group_abstraction_rtos_queue
+ * \{
+ */
+
+/** Create a queue.
+ *
+ * This is a queue of data where entries are placed on the back of the queue
+ * and removed from the front of the queue.
+ *
+ * length * itemsize must be a power of 2 and >= 2
```
ziemleszcz: Trailing space.

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
julianuziemblo: right, my bad...

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,911 @@
...
+            cdc_header_t* cdc_header = (cdc_header_t*)(data + sdpcm_header_size);
+            uint32_t cdc_size = size - sdpcm_header_size;
+
+            /* Send control request */
+            CHECK_RETURN(whd_bus_usb_send_ctrl(whd_driver, cdc_header, &cdc_size));
+
+            /* Receive control response */
+            status = whd_bus_usb_receive_ctrl_buffer(whd_driver, &rec_buffer);
+            if (status != WHD_SUCCESS && status != WHD_BUFFER_ALLOC_FAIL) {
+                whd_buffer_release(whd_driver, rec_buffer, WHD_NETWORK_RX);
```
ziemleszcz: I would still prefer to free this buffer in `whd_bus_usb_receive_ctrl_buffer()`.

ziemleszcz: `whd_buffer_release(whd_driver, buffer, WHD_NETWORK_TX)` is called in this path.

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,911 @@
...
+            }
+            else {
+                CHECK_RETURN(status);
```
ziemleszcz: `whd_buffer_release(whd_driver, buffer, WHD_NETWORK_TX)` is not called in this code path if status is error.

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,911 @@
...
+static whd_result_t whd_bus_usb_stop_rx(whd_driver_t whd_driver)
+{
+    cy_rslt_t result = CY_RSLT_SUCCESS;
+    whd_bus_set_state(whd_driver, WHD_FALSE);
+    whd_usb_rx_thread_notify();
+    whd_thread_notify(whd_driver);
+    if (whd_bus_usb_common.rx.thread != 0) {
+        result |= cy_rtos_join_thread(&whd_bus_usb_common.rx.thread);
+    }
+    result |= cy_rtos_queue_deinit(&whd_bus_usb_common.rx.queue);
```
ziemleszcz, line 425: Unconditional deinit of queue and semaphore.

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,911 @@
...
+}
+
+
+/***************************************************************************************************
+ * whd_bus_usb_init
+ **************************************************************************************************/
+whd_result_t whd_bus_usb_init(whd_driver_t whd_driver)
+{
+    whd_result_t result;
+
```
ziemleszcz: Spaces in blank line.

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
julianuziemblo: is there a way to check for it though? and is it such a problem we unconditionally deinit them?

ziemleszcz: These resources may be uninitialized in the error path.

julianuziemblo: yeah, but using free on NULL data in queue or resourceDestroy on improper mutex and cond doesn't do anything

ziemleszcz: Not true for mutex. Also, the queue or semaphore implementation may change in the future. I know it’s a bit of a pain, but resources should be freed properly.

julianuziemblo: I'll implement the checks for all of them in cyabs_rtos, as that will be more proper for if the implementation changes

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,907 @@
...
+                (unsigned short)(WHD_USB_MAX_RECEIVE_BUF_SIZE + (uint16_t)sizeof(whd_buffer_header_t)),
+                (whd_sdpcm_has_tx_packet(whd_driver) ? 0 : WHD_RX_BUF_TIMEOUT)) != WHD_SUCCESS) {
+        return WHD_BUFFER_ALLOC_FAIL;
+    }
+
+    void *data = whd_buffer_get_current_piece_data_pointer(whd_driver, *buffer);
+    size_t buflen = WHD_USB_MAX_RECEIVE_BUF_SIZE;
+    cy_rslt_t result = cyhal_usb_ctrl_receive(whd_driver->bus_priv->usb_obj, data, &buflen);
+
+    if (result != WHD_SUCCESS) {
```
ziemleszcz: CY_RSLT_SUCCESS

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,907 @@
...
+        whd_buffer_release(whd_driver, buffer, WHD_NETWORK_RX);
+    }
```
ziemleszcz, line 768: return WHD_HAL_ERROR

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,907 @@
...
+
+    return (result == CY_RSLT_SUCCESS) ? WHD_SUCCESS : WHD_HAL_ERROR;
```
ziemleszcz: return WHD_SUCCESS

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,907 @@
...
+            whd_process_cdc(whd_driver, rec_buffer);
+            break;
+        }
+
+        default:
+            whd_minor_assert("whd_bus_usb_send_buffer: SDPCM packet of unknown channel received - dropping packet", 0 != 0);
+            break;
+    }
+
+    whd_buffer_release(whd_driver, buffer, WHD_NETWORK_TX);
```
ziemleszcz, line 697: The buffer is still not released on `CHECK_RETURN` paths. I’m not sure how this is intended to work for USB, but in SDIO implementation the buffer is always freed (see `whd_bus_sdio_send_buffer`).

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,907 @@
...
+                (whd_sdpcm_has_tx_packet(whd_driver) ? 0 : WHD_RX_BUF_TIMEOUT)) != WHD_SUCCESS) {
+        return WHD_BUFFER_ALLOC_FAIL;
+    }
+
+    void *data = whd_buffer_get_current_piece_data_pointer(whd_driver, *buffer);
+    size_t buflen = WHD_USB_MAX_RECEIVE_BUF_SIZE;
+    cy_rslt_t result = cyhal_usb_ctrl_receive(whd_driver->bus_priv->usb_obj, data, &buflen);
+
+    if (result != WHD_SUCCESS) {
+        whd_buffer_release(whd_driver, buffer, WHD_NETWORK_RX);
```
ziemleszcz: buffer -> *buffer?

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,907 @@
...
+            /* Send control request */
+            CHECK_RETURN(whd_bus_usb_send_ctrl(whd_driver, cdc_header, &cdc_size));
+
+            /* Receive control response */
+            CHECK_RETURN(whd_bus_usb_receive_ctrl_buffer(whd_driver, &rec_buffer));
+
+            (void)whd_buffer_set_size(whd_driver, rec_buffer, cdc_size - 4);
+
+            /* Process CDC data... */
+            whd_process_cdc(whd_driver, rec_buffer);
```
ziemleszcz, line 688: Does `whd_process_cdc()` always free `rec_buffer`?

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,907 @@
...
+
+    result = whd_bus_usb_init_internal(whd_driver);
+    if (result != WHD_SUCCESS) {
+        cyhal_usb_free(whd_driver->bus_priv->usb_obj);
+        return result;
+    }
+
+    result = whd_bus_usb_start_rx(whd_driver);
+    if (result != WHD_SUCCESS) {
+        cyhal_usb_free(whd_driver->bus_priv->usb_obj);
```
ziemleszcz, line 274: The bus state remains UP on this path. A subsequent call to `whd_bus_usb_deinit()` will invoke `cyhal_usb_free()` a second time.

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,907 @@
...
+        if (result != WHD_SUCCESS) {
+            WPRINT_WHD_ERROR(("Failed to download firmware.\n"));
+            return result;
+        }
+
+        /* re-initialize USB as it's re-enumerated */
+        rslt = cyhal_usb_wait_state(whd_bus_usb_common.config.path, false, 5000 /* ms */);
+        if (rslt == CY_RTOS_TIMEOUT) {
+            WPRINT_WHD_ERROR(("Timed out waiting for USB to go down\n"));
+            return result;
```
ziemleszcz: Returns `WHD_SUCCESS`.

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,907 @@
...
+
+
+static whd_result_t whd_bus_usb_stop_rx(whd_driver_t whd_driver)
+{
+    cy_rslt_t result = CY_RSLT_SUCCESS;
+    whd_bus_set_state(whd_driver, WHD_FALSE);
+    whd_usb_rx_thread_notify();
+    whd_thread_notify(whd_driver);
+    result |= cy_rtos_join_thread(&whd_bus_usb_common.rx.thread);
+    result |= cy_rtos_queue_deinit(&whd_bus_usb_common.rx.queue);
```
ziemleszcz, line 425: Are you sure that all WHD buffers in the queue have been released at this point?

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
julianuziemblo: it seems so - if not in the function itself, then in WHD

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,925 @@
...
+
+    return WHD_SUCCESS;
+}
+
+
+static void whd_bus_usb_rx_queue_buffers_release(whd_driver_t whd_driver)
+{
+    whd_buffer_t buffer;
+    cy_rslt_t result = cy_rtos_queue_get(&whd_bus_usb_common.rx.queue, &buffer);
+    while (result == WHD_SUCCESS) {
```
ziemleszcz, line 410: `CY_RSLT_SUCCESS` or use `whd_bus_usb_rx_queue_dequeue()`.

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,925 @@
...
+                                           /* size   */ WHD_USB_TRX_RDL_CHUNK,
+                                           /* size   */ &size,
+                                           /* buffer */ buff));
+
+            /* Simply avoid having to send a ZLP by ensuring we never have an even multiple of 64 */
+            if (size == 0) {
+                /* nothing to do, we've reached the end */
+                break;
+            }
+
```
ziemleszcz: Trailing spaces in this line.

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,925 @@
...
+        if (rslt == CY_RTOS_TIMEOUT) {
+            WPRINT_WHD_ERROR(("Timed out waiting for USB to go down\n"));
+            return WHD_TIMEOUT;
+        }
+
+        (void)cyhal_usb_free(whd_driver->bus_priv->usb_obj);
+        result = whd_bus_usb_wait_and_init_device(whd_driver);
+        if (result != WHD_SUCCESS) {
+            WPRINT_WHD_ERROR(("Failed to wait for device.\n"));
+            return result;
```
ziemleszcz, line 245: `whd_usb_bus_init()` will call `cyhal_usb_free()` second time in this path.

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,925 @@
...
+{
+    cy_rslt_t result;
+    uint8_t *data = whd_buffer_get_current_piece_data_pointer(whd_driver, buffer);
+    size_t size = WHD_USB_MAX_RECEIVE_BUF_SIZE;
+
+    result = cyhal_usb_bulk_receive(whd_driver->bus_priv->usb_obj, data, &size);
+    if (result == CY_RSLT_SUCCESS) {
+        /* Set final size of received packet */
+        whd_buffer_set_size(whd_driver, buffer, size);
+        (void)whd_bus_usb_rx_queue_enqueue(&buffer);
```
ziemleszcz: Enqueue fail is ignored. The buffer will stay unused forever.

julianuziemblo: technically, we can't enter this path because we know by this point the queue is not full. But you're right this is better changed for futureproofing.

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -0,0 +1,924 @@
...
+}
+
+
+static void whd_bus_usb_rx_queue_buffers_release(whd_driver_t whd_driver)
+{
+    whd_buffer_t buffer;
+    whd_result_t result = whd_bus_usb_rx_queue_dequeue(&buffer);
+    while (result == WHD_SUCCESS) {
+        whd_buffer_release(whd_driver, buffer, WHD_NETWORK_RX);
+        result = cy_rtos_queue_get(&whd_bus_usb_common.rx.queue, &buffer);
```
ziemleszcz: whd_bus_usb_rx_queue_dequeue

ziemleszcz (approved):

PR #129, opened 2025-12-10
wi-fi: support STA mode

julianuziemblo:
## Description
This PR adds STA mode to Wi-Fi manager. For easier Wi-Fi management, the `/dev/wifi` device had to be split into three devices:
- `/dev/wifi/ctrl`: Wi-Fi state management: on/off. Putting Wi-Fi "on" initializes the WHD interfaces: STA as primary and AP as secondary. It also creates the latter 2 devices for managing STA and AP modes.
- `/dev/wifi/sta`: STA mode management: set idle timeout, set SSID and key of the network you want to connect to, connect/disconnect.
- `/dev/wifi/ap`: AP mode management: set idle timeout, set SSID and key of the access point, start/stop (API unchanged)

Idle timeout should always be set `!= 0` in both modes as it is used to periodically check if the bus is still operational (the bus can fail if e.g. device gets disconnected).

This PR also adds by default, concurrent mode (AP+STA together at the same time) is enabled. To disable it, define WIFI_CONCURRENT_MODE_ENABLED to 0 in `lwipopts.h`.

This PR also allows setting AP static address by writing to `/dev/wifi/ap`:
- "ip_addr a.b.c.d" for IP address
- "netmask a.b.c.d" for netmask
- "gateway a.b.c.d" for gateway address
Written address is only validated if it itself is valid. It is
not checked for being a valid interface/netmask/gateway address.
This check is only done on AP start.

This PR also also adds IP reporting on read:
- AP reports its static IP, netmask, gateway
- STA reports its dynamic IP, netmask, gateway

## Motivation and Context
STA mode (client mode) needed on some platforms (especially imxrt).

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change):
    - `/dev/wifi` device split into three devices (described in "Description" section above)

## How Has This Been Tested?

- [x] Tested by hand on: `armv7m7-imxrt117x-evk`,  `armv7a7-imx6ull-dataro`, NILEE.

## Checklist:

## Special treatment

    - https://github.com/phoenix-rtos/phoenix-rtos-lwip/pull/118 - *maybe not necessary, but needed to run on imrt117x with Sterling LWB5+ USB dongle*
    - https://github.com/phoenix-rtos/libphoenix/pull/466

### DIFF
--- a/wi-fi/lwip/main.c
+++ b/wi-fi/lwip/main.c
@@ -225,6 +225,56 @@ static void wifi_ap_main_loop(void)
 }
 
 
+static void wifi_deinit_interface(void)
+{
+	/* All below functions are safe to call even if interface has not been initialized */
+	(void)cy_lwip_network_down(&wifi_common.iface);
+	(void)cy_lwip_remove_interface(&wifi_common.iface);
+	cybsp_wifi_deinit(wifi_common.iface.whd_iface);
+	cybsp_free();
+}
+
+
+static int wifi_init_interface(cy_lwip_nw_interface_role_t role)
+{
+	cy_rslt_t result;
+
+	result = cybsp_init();
+	if (result != CY_RSLT_SUCCESS) {
+		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't init Wi-Fi HW\n");
+		return -1;
+	}
+
+	result = cybsp_wifi_init_primary(&wifi_common.iface.whd_iface);
+	if (result != CY_RSLT_SUCCESS) {
+		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't init Wi-Fi interface\n");
+		cybsp_free();
+		return -1;
+	}
+
+	wifi_common.iface.role = role;
+
+	result = cy_lwip_add_interface(&wifi_common.iface, &ap_addr);
+	if (result != CY_RSLT_SUCCESS) {
+		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't add Wi-Fi interface\n");
+		cybsp_wifi_deinit(wifi_common.iface.whd_iface);
+		cybsp_free();
+		return -1;
+	}
+
+	result = cy_lwip_network_up(&wifi_common.iface);
+	if (result != CY_RSLT_SUCCESS) {
+		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't bring up Wi-Fi interface\n");
+		cy_lwip_remove_interface(&wifi_common.iface);
+		cybsp_wifi_deinit(wifi_common.iface.whd_iface);
+		cybsp_free();
+		return -1;
+	}
+
+	return 0;
+}
+
+
 static void wifi_ap_thread(void *arg)
 {
 	bool started = false;
@@ -243,55 +293,21 @@ static void wifi_ap_thread(void *arg)
 
 		mutexUnlock(wifi_common.lock);
 
-		result = cybsp_init();
-		if (result != CY_RSLT_SUCCESS) {
-			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't init Wi-Fi HW\n");
-			break;
-		}
-
-		result = cybsp_wifi_init_primary(&wifi_common.iface.whd_iface);
-		if (result != CY_RSLT_SUCCESS) {
-			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't init Wi-Fi interface\n");
-			cybsp_free();
-			break;
-		}
-
-		wifi_common.iface.role = CY_LWIP_AP_NW_INTERFACE;
-
-		result = cy_lwip_add_interface(&wifi_common.iface, &ap_addr);
-		if (result != CY_RSLT_SUCCESS) {
-			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't add Wi-Fi interface\n");
-			cybsp_wifi_deinit(wifi_common.iface.whd_iface);
-			cybsp_free();
-			break;
-		}
-
-		result = cy_lwip_network_up(&wifi_common.iface);
-		if (result != CY_RSLT_SUCCESS) {
-			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't bring up Wi-Fi interface\n");
-			cy_lwip_remove_interface(&wifi_common.iface);
-			cybsp_wifi_deinit(wifi_common.iface.whd_iface);
-			cybsp_free();
+		if (wifi_init_interface(CY_LWIP_AP_NW_INTERFACE) < 0) {
 			break;
 		}
 
 		result = whd_wifi_init_ap(wifi_common.iface.whd_iface, &ssid, AP_SECURITY_MODE, key, key_len, AP_CHANNEL);
 		if (result != WHD_SUCCESS) {
 			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't init Wi-Fi AP\n");
-			cy_lwip_network_down(&wifi_common.iface);
-			cy_lwip_remove_interface(&wifi_common.iface);
-			cybsp_wifi_deinit(wifi_common.iface.whd_iface);
-			cybsp_free();
+			wifi_deinit_interface();
 			break;
 		}
 
 		result = whd_wifi_start_ap(wifi_common.iface.whd_iface);
 		if (result != WHD_SUCCESS) {
 			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't start Wi-Fi AP\n");
-			cy_lwip_network_down(&wifi_common.iface);
-			cy_lwip_remove_interface(&wifi_common.iface);
-			cybsp_wifi_deinit(wifi_common.iface.whd_iface);
-			cybsp_free();
+			wifi_deinit_interface();
 			break;
 		}
 
@@ -307,10 +323,7 @@ static void wifi_ap_thread(void *arg)
 		wifi_ap_main_loop();
 
 		whd_wifi_stop_ap(wifi_common.iface.whd_iface);
-		cy_lwip_network_down(&wifi_common.iface);
-		cy_lwip_remove_interface(&wifi_common.iface);
-		cybsp_wifi_deinit(wifi_common.iface.whd_iface);
-		cybsp_free();
+		wifi_deinit_interface();
 	}
 
 	mutexLock(wifi_common.lock);
--- a/wi-fi/lwip/main.c
+++ b/wi-fi/lwip/main.c
@@ -27,9 +27,9 @@
 #define WIFI_THREAD_PRIO    4
 #define WIFI_THREAD_STACKSZ (4 * _PAGE_SIZE)
 
-#define WIFI_FLAG_STARTED (1 << 0)
-#define WIFI_FLAG_FAILED  (1 << 1)
-#define WIFI_FLAG_FINISH  (1 << 2)
+#define WIFI_FLAG_STARTED (1U << 0)
+#define WIFI_FLAG_FAILED  (1U << 1)
+#define WIFI_FLAG_FINISH  (1U << 2)
 
 #define WIFI_START_RETRIES 5
 
@@ -39,12 +39,12 @@
 #define AP_SECURITY_MODE WHD_SECURITY_WPA2_AES_PSK
 #define AP_CHANNEL       1
 
-#define SNPRINTF_APPEND(fmt, ...) \
+#define SNPRINTF_APPEND(overflow, fmt, ...) \
 	do { \
 		if (!overflow) { \
 			int n = snprintf(buf, size, fmt, ##__VA_ARGS__); \
 			if (n >= size) \
-				overflow = 1; \
+				overflow = true; \
 			else { \
 				size -= n; \
 				buf += n; \
@@ -68,7 +68,7 @@ static struct {
 	cy_lwip_nw_interface_t iface;
 
 	struct {
-		int busy;
+		bool busy;
 		char buf[128];
 		int len;
 	} dev;
@@ -85,6 +85,7 @@ static int wifi_ap_set_idle_timeout(const char *data, size_t len)
 {
 	char buf[16];
 	long int timeout;
+	char *endp;
 
 	if (len > (sizeof(buf) - 1)) {
 		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't set Wi-Fi idle timeout (max length is %u)\n", sizeof(buf) - 1);
@@ -94,14 +95,22 @@ static int wifi_ap_set_idle_timeout(const char *data, size_t len)
 	memcpy(buf, data, len);
 	buf[len] = '\0';
 
-	timeout = strtol(buf, NULL, 0);
-	if (timeout < 0) {
+	errno = 0;
+	timeout = strtol(buf, &endp, 0);
+	if (errno != 0 || endp == buf) {
+		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't set Wi-Fi idle timeout (bad timeout value)\n");
+		return -1;
+	}
+	else if (timeout < 0) {
 		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't set Wi-Fi idle timeout (min value is 0)\n");
 		return -1;
 	}
 	else if (timeout > UINT32_MAX) {
 		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't set Wi-Fi idle timeout (max value is %u)\n", UINT32_MAX);
 		return -1;
+	}
+	else {
+		/* nothing */
 	}
 
 	mutexLock(wifi_common.lock);
@@ -149,7 +158,7 @@ static int wifi_ap_set_key(const char *key, size_t len)
 }
 
 
-static int wifi_ap_is_idle(void)
+static bool wifi_ap_is_idle(void)
 {
 	uint8_t buf[sizeof(uint32_t) + 4 * sizeof(whd_mac_t)];
 	whd_maclist_t *clients = (whd_maclist_t *)buf;
@@ -159,10 +168,11 @@ static int wifi_ap_is_idle(void)
 	clients->count = 4;
 
 	result = whd_wifi_get_associated_client_list(wifi_common.iface.whd_iface, buf, sizeof(buf));
-	if (result == WHD_SUCCESS && clients->count == 0)
-		return 1;
+	if (result == WHD_SUCCESS && clients->count == 0) {
+		return true;
+	}
 
-	return 0;
+	return false;
 }
 
 
@@ -176,8 +186,9 @@ static void wifi_ap_main_loop(void)
 
 		mutexLock(wifi_common.lock);
 
-		if (wifi_common.flags & WIFI_FLAG_FINISH)
+		if ((wifi_common.flags & WIFI_FLAG_FINISH) != 0) {
 			finish = true;
+		}
 
 		if (wifi_common.idle_timeout != 0) {
 			if (wifi_common.idle_current < wifi_common.idle_timeout) {
@@ -193,18 +204,21 @@ static void wifi_ap_main_loop(void)
 		mutexUnlock(wifi_common.lock);
 
 		if (!finish && check_idle) {
-			if (wifi_ap_is_idle())
+			if (wifi_ap_is_idle()) {
 				finish = true;
+			}
 		}
 
-		if (finish)
+		if (finish) {
 			break;
+		}
 
 		mutexLock(wifi_common.lock);
 		condWait(wifi_common.cond, wifi_common.lock, cond_timeout * 1000000ULL);
 
-		if (update_idle)
+		if (update_idle) {
 			wifi_common.idle_current += 1;
+		}
 
 		mutexUnlock(wifi_common.lock);
 	}
@@ -335,7 +349,7 @@ static int wifi_ap_start(void)
 
 	mutexUnlock(wifi_common.lock);
 
-	return (flags & WIFI_FLAG_STARTED ? 0 : -1);
+	return ((flags & WIFI_FLAG_STARTED) != 0) ? 0 : -1;
 }
 
 
@@ -364,26 +378,28 @@ static int wifi_dev_open(int flags)
 {
 	char *buf;
 	size_t size;
-	int overflow = 0;
+	bool overflow = false;
 
-	if (wifi_common.dev.busy)
+	if (wifi_common.dev.busy) {
 		return -EBUSY;
+	}
 
 	buf = wifi_common.dev.buf;
 	size = sizeof(wifi_common.dev.buf);
 
 	mutexLock(wifi_common.lock);
 
-	SNPRINTF_APPEND("running=%u\n", wifi_common.tid != 0);
-	SNPRINTF_APPEND("timeout=%u\n", wifi_common.idle_timeout);
-	SNPRINTF_APPEND("ssid=%.*s\n", wifi_common.ssid.length, wifi_common.ssid.value);
+	SNPRINTF_APPEND(overflow, "running=%u\n", wifi_common.tid != 0);
+	SNPRINTF_APPEND(overflow, "timeout=%u\n", wifi_common.idle_timeout);
+	SNPRINTF_APPEND(overflow, "ssid=%.*s\n", wifi_common.ssid.length, wifi_common.ssid.value);
 
 	mutexUnlock(wifi_common.lock);
 
-	if (overflow)
+	if (overflow) {
 		return -EFBIG;
+	}
 
-	wifi_common.dev.busy = 1;
+	wifi_common.dev.busy = true;
 	wifi_common.dev.len = buf - wifi_common.dev.buf;
 
 	return 0;
@@ -392,9 +408,10 @@ static int wifi_dev_open(int flags)
 
 static int wifi_dev_close(void)
 {
-	if (!wifi_common.dev.busy)
+	if (!wifi_common.dev.busy) {
 		return -EBADF;
-	wifi_common.dev.busy = 0;
+	}
+	wifi_common.dev.busy = false;
 	return 0;
 }
 
@@ -403,8 +420,9 @@ static int wifi_dev_read(char *data, size_t size, off_t offset)
 {
 	int cnt;
 
-	if (offset > wifi_common.dev.len)
+	if (offset > wifi_common.dev.len) {
 		return -ERANGE;
+	}
 
 	cnt = min(size, wifi_common.dev.len - offset);
 	memcpy(data, wifi_common.dev.buf + offset, cnt);
@@ -436,7 +454,7 @@ static int wifi_dev_write(const char *data, size_t size)
 		wifi_ap_stop();
 	}
 	else {
-		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "got unknown Wi-Fi command\n");
+		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "got unknown Wi-Fi command: %.*s\n", (int)size, data);
 		return -EINVAL;
 	}
 
@@ -459,7 +477,7 @@ static void wifi_msg_thread(void *arg)
 
 	result = cy_log_init(CY_LOG_INFO);
 	if (result != CY_RSLT_SUCCESS) {
-		printf("phoenix-rtos-lwip: can't init Wi-Fi logs\n");
+		fprintf(stderr, "phoenix-rtos-lwip: can't init Wi-Fi logs\n");
 		return;
 	}
 
@@ -477,8 +495,9 @@ static void wifi_msg_thread(void *arg)
 		msg_t msg = { 0 };
 		msg_rid_t rid;
 
-		if (msgRecv(port, &msg, &rid) < 0)
+		if (msgRecv(port, &msg, &rid) < 0) {
 			continue;
+		}
 
 		switch (msg.type) {
 			case mtOpen:
@@ -512,17 +531,18 @@ __constructor__(1000) void init_wifi(void)
 	int err;
 
 	err = mutexCreate(&wifi_common.lock);
-	if (err) {
+	if (err < 0) {
 		errout(err, "mutexCreate(lock)");
 	}
 
 	err = condCreate(&wifi_common.cond);
-	if (err) {
+	if (err < 0) {
 		resourceDestroy(wifi_common.lock);
 		errout(err, "condCreate(cond)");
 	}
 
-	if ((err = sys_thread_opt_new("wifi-msg", wifi_msg_thread, NULL, WIFI_THREAD_STACKSZ, WIFI_THREAD_PRIO, NULL))) {
+	err = sys_thread_opt_new("wifi-msg", wifi_msg_thread, NULL, WIFI_THREAD_STACKSZ, WIFI_THREAD_PRIO, NULL);
+	if (err < 0) {
 		resourceDestroy(wifi_common.lock);
 		resourceDestroy(wifi_common.cond);
 		errout(err, "thread(wifi-msg)");
--- a/wi-fi/lwip/main.c
+++ b/wi-fi/lwip/main.c
@@ -3,48 +3,74 @@
  *
  * LwIP Wi-Fi
  *
- * Copyright 2021 Phoenix Systems
- * Author: Ziemowit Leszczynski
+ * Copyright 2021, 2026 Phoenix Systems
+ * Author: Ziemowit Leszczynski, Julian Uziembło
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-#include "lwip/sys.h"
+#include <ctype.h>
+#include <string.h>
+#include <sys/minmax.h>
+#include <sys/msg.h>
+#include <sys/stat.h>
+#include <posix/utils.h>
 
 #include "whd_wifi_api.h"
 #include "whd_wlioctl.h"
 #include "cybsp.h"
 #include "cybsp_wifi.h"
 #include "cy_lwip.h"
-#include "cyabs_rtos.h"
 #include "cy_lwip_log.h"
+#include "cy_log.h"
 
-#include <string.h>
-#include <sys/minmax.h>
-#include <posix/utils.h>
+#include "lwipopts.h"
+#include "netif.h"
+#include "lwip/sys.h"
 
+/* timeout for STA scanning in seconds */
+#ifndef WIFI_STA_SCAN_TIMEOUT_SECONDS
+#define WIFI_STA_SCAN_TIMEOUT_SECONDS 15
+#endif
+
+/* number of STA scans in the timeout time */
+#ifndef WIFI_STA_N_SCANS
+#define WIFI_STA_N_SCANS 3
+#endif
+
+_Static_assert(WIFI_STA_SCAN_TIMEOUT_SECONDS > 0, "STA scan timeout has to be positive");
+_Static_assert(WIFI_STA_N_SCANS > 0, "Number of STA scans has to be positive");
+_Static_assert(((float)WIFI_STA_SCAN_TIMEOUT_SECONDS / WIFI_STA_N_SCANS) >= 5.0f, "At least 5 seconds needed for every STA scan");
 
 #define WIFI_THREAD_PRIO    4
-#define WIFI_THREAD_STACKSZ (4 * _PAGE_SIZE)
+#define WIFI_THREAD_STACKSZ 4096
 
 #define WIFI_FLAG_STARTED (1U << 0)
 #define WIFI_FLAG_FAILED  (1U << 1)
 #define WIFI_FLAG_FINISH  (1U << 2)
 
-#define WIFI_START_RETRIES 5
+#define STA_DEV_ID    0
+#define STA_DEV_NAME  "/dev/wifi/sta"
+#define STA_SCAN_TYPE WHD_SCAN_TYPE_ACTIVE
+#define STA_BSS_TYPE  WHD_BSS_TYPE_ANY
 
-#define WIFI_DEV_ID   0
-#define WIFI_DEV_NAME "/dev/wifi"
-
+#define AP_DEV_ID        1
+#define AP_DEV_NAME      "/dev/wifi/ap"
 #define AP_SECURITY_MODE WHD_SECURITY_WPA2_AES_PSK
 #define AP_CHANNEL       1
 
-#define SNPRINTF_APPEND(overflow, fmt, ...) \
+#define CTRL_DEV_ID   2
+#define CTRL_DEV_NAME "/dev/wifi/ctrl"
+
+#define CONST_STRLEN(str) (sizeof(str) - 1)
+
+#define SNPRINTF_APPEND(overflow, buf, size, fmt, ...) \
 	do { \
 		if (!overflow) { \
 			int n = snprintf(buf, size, fmt, ##__VA_ARGS__); \
-			if (n >= size) \
+			if (n >= size) { \
 				overflow = true; \
+			} \
 			else { \
 				size -= n; \
 				buf += n; \
@@ -53,72 +79,76 @@
 	} while (0)
 
 
-static struct {
+struct wifi_device {
 	handle_t lock;
 	handle_t cond;
-	handle_t tid;
 
-	volatile uint8_t flags;
-	uint32_t idle_timeout;
-	uint32_t idle_current;
 	whd_ssid_t ssid;
-	uint8_t key[WSEC_MAX_PSK_LEN];
-	uint8_t key_len;
+	struct {
+		uint8_t value[WSEC_MAX_PSK_LEN];
+		uint8_t len;
+	} key;
 
+
+	bool busy;
+	int len;
+	char buf[128];
 	cy_lwip_nw_interface_t iface;
 
-	struct {
-		bool busy;
-		char buf[128];
-		int len;
-	} dev;
-} wifi_common;
+	union {
+		struct {
+			handle_t tid;
+			volatile uint8_t flags;
+			uint32_t idleCurrent;
+			uint32_t timeout;
+			ip_static_addr_t addr;
+		} apPriv;
 
-static ip_static_addr_t ap_addr = {
-	.addr = IPADDR4_INIT_BYTES(192, 168, 2, 1),
-	.netmask = IPADDR4_INIT_BYTES(255, 255, 255, 0),
-	.gateway = IPADDR4_INIT_BYTES(192, 168, 2, 1)
+		struct {
+			handle_t tid;
+			whd_scan_result_t scanResult;
+		} staPriv;
+	};
 };
 
 
-static int wifi_ap_set_idle_timeout(const char *data, size_t len)
-{
-	char buf[16];
-	long int timeout;
-	char *endp;
+enum wifi_state {
+	wifi_off,
+	wifi_idle,
+	wifi_running,
+};
 
-	if (len > (sizeof(buf) - 1)) {
-		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't set Wi-Fi idle timeout (max length is %u)\n", sizeof(buf) - 1);
-		return -1;
-	}
 
-	memcpy(buf, data, len);
-	buf[len] = '\0';
+static struct {
+	bool initialized;
+	uint32_t msgport;
+	handle_t msgtid;
 
-	errno = 0;
-	timeout = strtol(buf, &endp, 0);
-	if (errno != 0 || endp == buf) {
-		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't set Wi-Fi idle timeout (bad timeout value)\n");
-		return -1;
-	}
-	else if (timeout < 0) {
-		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't set Wi-Fi idle timeout (min value is 0)\n");
-		return -1;
-	}
-	else if (timeout > UINT32_MAX) {
-		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't set Wi-Fi idle timeout (max value is %u)\n", UINT32_MAX);
-		return -1;
-	}
-	else {
-		/* nothing */
-	}
+	union {
+		struct {
+			struct wifi_device sta;
+			struct wifi_device ap;
+		};
+		struct wifi_device dev[2];
+	};
 
-	mutexLock(wifi_common.lock);
-	wifi_common.idle_timeout = timeout;
-	mutexUnlock(wifi_common.lock);
-
-	return 0;
-}
+	struct {
+		bool busy;
+		uint32_t port;
+		int len;
+		char buf[16];
+		handle_t lock;
+		handle_t cond;
+		handle_t tid;
+	} ctrl;
+} wifi_common = {
+	.ap.apPriv.addr = {
+		/* default AP address */
+		.addr = IPADDR4_INIT_BYTES(192, 168, 2, 1),
+		.netmask = IPADDR4_INIT_BYTES(255, 255, 255, 0),
+		.gateway = IPADDR4_INIT_BYTES(192, 168, 2, 1),
+	},
+};
 
 
 static const char *trim(const char *str, size_t *len)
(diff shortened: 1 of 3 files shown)


wi-fi/lwip/main.c
```diff
@@ -221,234 +316,576 @@ static void wifi_ap_thread(void *arg)
...
+		default:
+			return;
+	}
+
+	mutexLock(wifi_common.dev[STA_DEV_ID].lock);
+	condSignal(wifi_common.dev[STA_DEV_ID].cond);
+	if (result_ptr != NULL) {
+		*result_ptr = NULL;
+	}
+	mutexUnlock(wifi_common.dev[STA_DEV_ID].lock);
```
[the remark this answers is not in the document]
julianuziemblo: WHD internally passes &scan_result as pointer, so it IS passed as `&(&scan_result)` anyway. Making it NULL just signals to WHD to stop scanning as we've encountered the appropriate result.

wi-fi/lwip/main.c
```diff
@@ -94,15 +95,23 @@ static int wifi_ap_set_idle_timeout(const char *data, size_t len)
 	memcpy(buf, data, len);
 	buf[len] = '\0';
 
-	timeout = strtol(buf, NULL, 0);
-	if (timeout < 0) {
+	errno = 0;
+	timeout = strtol(buf, &endp, 0);
+	if (errno != 0 || endp == buf || timeout == LONG_MAX) {
```
ziemleszcz: `timeout == LONG_MAX` should be already handled by `errno != 0` (ERANGE).

wi-fi/lwip/main.c
```diff
@@ -121,13 +121,33 @@ static int wifi_ap_set_idle_timeout(const char *data, size_t len)
...
+
+
 static int wifi_ap_set_ssid(const char *ssid, size_t len)
 {
 	if (len > SSID_NAME_SIZE) {
 		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't set Wi-Fi SSID (max length is %u)\n", SSID_NAME_SIZE);
 		return -1;
 	}
 
+	ssid = trim(ssid, &len);
```
ziemleszcz: Move it before SSID size check.

wi-fi/lwip/main.c
```diff
@@ -149,6 +169,8 @@ static int wifi_ap_set_key(const char *key, size_t len)
 		return -1;
 	}
 
+	key = trim(key, &len);
```
ziemleszcz: Move it before PSK size checks.

wi-fi/lwip/main.c
```diff
@@ -446,6 +468,7 @@ static int wifi_dev_read(char *data, size_t size, off_t offset)
 
 static int wifi_dev_write(const char *data, size_t size)
 {
+	data = trim(data, &size);
```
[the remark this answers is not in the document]
julianuziemblo: I don't think it will as before `strncmp` there is `size >= 8` check which will not be passed, hence not calling `wifi_ap_set_idle_timeout`

wi-fi/lwip/main.c
```diff
@@ -53,25 +65,42 @@
...
+			struct wifi_device sta;
+			struct wifi_device ap;
+		};
+		struct wifi_device dev[2];
+	};
 
 	struct {
 		bool busy;
-		char buf[128];
+		char buf[16];
```
ziemleszcz: You can use move it to stack since it's very small.

wi-fi/lwip/main.c
```diff
@@ -53,25 +65,41 @@
...
+
+	volatile uint8_t flags;
+	uint32_t idle_current;
+
+	union {
+		struct {
+			struct wifi_device sta;
+			struct wifi_device ap;
+		};
+		struct wifi_device dev[2];
```
ziemleszcz: Assuming that `CY_LWIP_AP_NW_INTERFACE` is 0 and `CY_LWIP_STA_NW_INTERFACE` is 1 is dangerous.

wi-fi/lwip/main.c
```diff
@@ -158,29 +219,31 @@ static int wifi_ap_is_idle(void)
...
+		return true;
+	}
 
-	return 0;
+	return false;
 }
 
 
-static void wifi_ap_main_loop(void)
+static void wifi_ap_main_loop(void)
```
ziemleszcz: Remove extra space at the end of line.

wi-fi/lwip/main.c
```diff
@@ -53,25 +65,41 @@
...
+	char buf[128];
 	cy_lwip_nw_interface_t iface;
+};
+
+
+static struct {
+	atomic_bool initialized;
+	handle_t tid;
+
+	volatile uint8_t flags;
```
ziemleszcz: `flags` is used by AP only and protected by `ap.lock` so you should move it to `wifi_device` struct.

wi-fi/lwip/main.c
```diff
@@ -221,234 +315,583 @@ static void wifi_ap_thread(void *arg)
...
+		wifi_set_key(id, data + 4, size - 4);
 	}
-	else if (strncmp("start", data, size) == 0) {
-		unsigned int retries = WIFI_START_RETRIES;
-
-		while (wifi_ap_start() < 0 && retries-- > 0) {
-			/* FIXME: temporary workaround - find out why AP doesn't start */
-			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "retrying to start Wi-Fi AP\n");
+	else if (id == STA_DEV_ID) {
+		if (size == (sizeof("connect") - 1) && strncmp("connect", data, size) == 0) {
```
ziemleszcz: Consider moving `id == STA_DEV_ID` and `id == AP_DEV_ID` checks to command `if` condition to avoid repeating error message.

wi-fi/lwip/main.c
julianuziemblo: oh yeah, I forgot it matters here, my bad.

wi-fi/lwip/main.c
```diff
@@ -221,234 +315,583 @@ static void wifi_ap_thread(void *arg)
...
+static int wifi_sta_connect(void)
+{
+	whd_result_t result;
+	whd_ssid_t ssid, connected_to;
+	uint8_t key[WSEC_MAX_PSK_LEN];
+	uint8_t key_len;
+	time_t when, now, timeout_remaining;
+	uint32_t scan_timeout;
+
+	mutexLock(wifi_common.sta.lock);
```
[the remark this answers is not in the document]
julianuziemblo: I now refactored the STA to work in similar fashion to AP: it spawns its own thread and periodically calls "wifi_is_idle" (`wifi_sta_is_connected` in this case) to check if the connection is still present. All checks are now performed the same for both wifi devices in `wifi_dev_start()` and `wifi_dev_stop()`.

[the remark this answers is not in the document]
julianuziemblo: I also added `!wifi_common.initialized` check to the msgthread

wi-fi/lwip/main.c
```diff
@@ -221,234 +357,590 @@ static void wifi_ap_thread(void *arg)
...
-		wifi_ap_set_ssid(data + 5, size - 5);
+		wifi_set_ssid(id, data + 5, size - 5);
 	}
 	else if (size >= 4 && strncmp("key ", data, 4) == 0) {
-		wifi_ap_set_key(data + 4, size - 4);
+		wifi_set_key(id, data + 4, size - 4);
 	}
-	else if (strncmp("start", data, size) == 0) {
-		unsigned int retries = WIFI_START_RETRIES;
+	else if (id == STA_DEV_ID && size == (sizeof("connect") - 1) && strncmp("connect", data, size) == 0) {
```
ziemleszcz: Consider adding `STRLEN` macro `(sizeof(s) - 1)`, since this pattern is repeated several times.

wi-fi/lwip/main.c
```diff
@@ -221,234 +357,590 @@ static void wifi_ap_thread(void *arg)
...
+		mutexLock(wifi_common.sta.lock);
+		do {
+			result = whd_wifi_scan(wifi_common.sta.iface.whd_iface, STA_SCAN_TYPE, STA_BSS_TYPE,
+					&ssid, NULL, NULL, NULL, wifi_scan_result_cb, &scan_result, &ssid);
+			condWait(wifi_common.sta.cond, wifi_common.sta.lock, timeout_remaining * 1000000ULL / STA_N_SCANS);
+
+			(void)whd_wifi_stop_scan(wifi_common.sta.iface.whd_iface);
+
+			if (scan_timeout != 0) {
+				now = time(NULL);
```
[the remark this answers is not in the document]
julianuziemblo: okay, good catch then - did not think about it

wi-fi/lwip/main.c
```diff
@@ -221,234 +357,590 @@ static void wifi_ap_thread(void *arg)
...
+
+		case mtWrite: {
+			const char *data = msg->i.data;
+			size_t size = msg->i.size;
+
+			data = trim(data, &size);
+
+			if (size >= (sizeof("on") - 1) && strncmp("on", data, size) == 0) {
+				oid_t oid __attribute__((unused));
+				if (wifi_common.initialized || lookup(AP_DEV_NAME, NULL, &oid) >= 0 || lookup(STA_DEV_NAME, NULL, &oid) >= 0) {
```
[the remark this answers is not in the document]
julianuziemblo: there's no way to wait at this point - maybe failing here is a good thing after all?

`sys_thread_join` does not work - as I said, we cannot wait due to sequential handling of messages. The best we can do is fail with a "too soon" message, as is done for now.

wi-fi/lwip/main.c
```diff
@@ -121,8 +121,28 @@ static int wifi_ap_set_idle_timeout(const char *data, size_t len)
 }
 
 
+static const char *trim(const char *str, size_t *len)
+{
+	size_t sz = *len;
+	while (sz > 0 && !isgraph(*str)) {
```
Darchiv: Unfortunately, `isgraph()` takes only ASCII printable characters into account and ignores locale. All Unicode characters (UTF-8 encoded as >=0x80) are trimmed. For future reference.

Is WHD/chip even capable of setting non-ASCII SSIDs?

wi-fi/lwip/main.c
```diff
@@ -457,9 +1028,9 @@ static void wifi_msg_thread(void *arg)
 	cy_rslt_t result;
 	unsigned int port;
 
-	result = cy_log_init(CY_LOG_INFO);
+	result = cy_log_init(CY_LOG_DEBUG);
```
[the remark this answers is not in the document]
julianuziemblo: uh, missed this one

wi-fi/lwip/main.c
```diff
@@ -140,315 +220,806 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+	if (clients->count == 0) {
+		return wifi_idle;
+	}
 
-	return 0;
+	return wifi_running;
+}
+
+
+static bool wifi_sta_state(whd_ssid_t *ssid)
```
Darchiv: Wrong return value. Should be `enum wifi_state`

wi-fi/lwip/main.c
```diff
@@ -80,56 +127,89 @@ static ip_static_addr_t ap_addr = {
 	.gateway = IPADDR4_INIT_BYTES(192, 168, 2, 1)
 };
 
+static whd_scan_result_t scanResult;
```
Darchiv: Why store it outside of `wifi_common`?

wi-fi/lwip/main.c
```diff
@@ -80,56 +127,89 @@ static ip_static_addr_t ap_addr = {
 	.gateway = IPADDR4_INIT_BYTES(192, 168, 2, 1)
 };
```
Darchiv, line 152: So AP network IPs are hardcoded? Optimally, these should be settable via `/dev/wifi/ap`. If out of scope, then please mark these with a TODO and `const` (if possible)

Darchiv (commented):
Left some remarks.

Is /dev/wifi/* interface documented somewhere? I mean, you have written what has changed in this PR, but is there a single point of reference for these devices?

wi-fi/lwip/main.c
```diff
@@ -140,315 +220,806 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+		cy_lwip_remove_interface(interface);
+		return -1;
+	}
 
 	return 0;
 }
 
 
-static int wifi_ap_is_idle(void)
+static bool wifi_ap_state(void)
```
Darchiv: Wrong return value. Should be `enum wifi_state`

julianuziemblo:
> Is /dev/wifi/* interface documented somewhere? I mean, you have written what has changed in this PR, but is there a single point of reference for these devices?

AFAIK: no, there isn't. Maybe I could create document on YT and then we could consider adding it to our documentation?

wi-fi/lwip/main.c
julianuziemblo: It's technically possible as `whd_ssid_t` is `length` and `value`, but I don't know if we wanna support it out-of-the-box. I'm also not sure if the chips support it. What do you think: should anything be done here?

wi-fi/lwip/main.c
julianuziemblo: mimicked `ap_addr`. I'll transfer it to common with `ap_addr` as we're making static address change possible.

wi-fi/lwip/main.c
julianuziemblo: how about now?

wi-fi/lwip/main.c
```diff
@@ -140,315 +222,827 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
 
-		result = whd_wifi_init_ap(wifi_common.iface.whd_iface, &ssid, AP_SECURITY_MODE, key, key_len, AP_CHANNEL);
 		if (result != WHD_SUCCESS) {
-			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't init Wi-Fi AP\n");
-			cy_lwip_network_down(&wifi_common.iface);
-			cy_lwip_remove_interface(&wifi_common.iface);
-			cybsp_wifi_deinit(wifi_common.iface.whd_iface);
-			cybsp_free();
+			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't get network's security\n");
 			break;
```
ziemleszcz, line 525: Unlocked `wifi_common.sta.lock`.

wi-fi/lwip/main.c
```diff
@@ -140,315 +222,827 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
-		result = whd_wifi_start_ap(wifi_common.iface.whd_iface);
+		result = whd_wifi_join_specific(wifi_common.sta.iface.whd_iface, &wifi_common.scanResult, key, keyLen);
 		if (result != WHD_SUCCESS) {
-			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't start Wi-Fi AP\n");
-			cy_lwip_network_down(&wifi_common.iface);
-			cy_lwip_remove_interface(&wifi_common.iface);
-			cybsp_wifi_deinit(wifi_common.iface.whd_iface);
-			cybsp_free();
+			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't join requested network\n");
 			break;
```
ziemleszcz, line 532: Unlocked `wifi_common.sta.lock`.

wi-fi/lwip/main.c
```diff
@@ -140,315 +222,827 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+
+	if (tid != 0) {
+		wifi_dev_stop(&wifi_common.sta);
+	}
+
+	if (wifi_common.sta.iface.whd_iface != NULL) {
+		cybsp_wifi_deinit(wifi_common.sta.iface.whd_iface);
+		cybsp_free();
+	}
+	else if (wifi_common.ap.iface.whd_iface != NULL) {
```
ziemleszcz, line 930: Why `else if` instead of `if`?

wi-fi/lwip/main.c
```diff
@@ -140,315 +222,827 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
 			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "retrying to start Wi-Fi AP\n");
 		}
 	}
-	else if (strncmp("stop", data, size) == 0) {
-		wifi_ap_stop();
+	else if (id == AP_DEV_ID && size == CONST_STRLEN("stop") && strncmp("stop", data, size) == 0) {
+		wifi_dev_stop(&wifi_common.ap);
+	}
+	else if (id == AP_DEV_ID && size >= CONST_STRLEN("ip_addr ") && strncmp("ip_addr ", data, 8) == 0) {
+		if (ip4addr_aton(data + 8, &wifi_common.apAddr.addr) != 0) {
```
ziemleszcz: Note that `data + 8` may be not NULL terminated string.

wi-fi/lwip/main.c
```diff
@@ -140,315 +222,827 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+
+		case mtClose:
+			wifi_common.ctrl.busy = false;
+			return EOK;
+
+		case mtRead: {
+			off_t offset = msg->i.io.offs;
+			int cnt = min(msg->o.size, wifi_common.ctrl.len - offset);
+
+			if (offset > wifi_common.ctrl.len) {
```
ziemleszcz: Move it before `cnt` calculation.

wi-fi/lwip/main.c
julianuziemblo: good eye, thx

wi-fi/lwip/main.c
julianuziemblo: cybsp_wifi_deinit() deinitializes both interfaces at once; the iface pointer is just to retrieve the driver. I can add comment here, it's commented like that in cybsp_wifi_deinit definition. What do you think?

wi-fi/lwip/main.c
julianuziemblo: Do you think we should terminate it here somehow, or just let `ip4addr_aton` do its thing?

wi-fi/lwip/main.c
ziemleszcz: Please add some comment.

wi-fi/lwip/main.c
```diff
@@ -140,315 +224,863 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
 			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "retrying to start Wi-Fi AP\n");
 		}
 	}
-	else if (strncmp("stop", data, size) == 0) {
-		wifi_ap_stop();
+	else if (id == AP_DEV_ID && size == CONST_STRLEN("stop") && strncmp("stop", data, size) == 0) {
+		wifi_dev_stop(&wifi_common.ap);
+	}
+	else if (id == AP_DEV_ID && size >= CONST_STRLEN("ip_addr ") && strncmp("ip_addr ", data, 8) == 0) {
+		return wifi_write_addr(&wifi_common.apAddr.addr, data + 8, size - 8);
```
ziemleszcz: I would follow `wifi_set_` name convention for these functions.

wi-fi/lwip/main.c
```diff
@@ -53,83 +74,146 @@
...
+	mutexUnlock(wifi_common.dev[id].lock);
 
 	return 0;
 }
 
 
-static int wifi_ap_set_ssid(const char *ssid, size_t len)
+static int wifi_set_ssid(id_t id, const char *ssid, size_t len)
 {
+	ssid = trim(ssid, &len);
```
ziemleszcz: Input is already trimmed.

wi-fi/lwip/main.c
```diff
@@ -53,83 +74,146 @@
...
+	mutexUnlock(wifi_common.dev[id].lock);
 
 	return 0;
 }
 
 
-static int wifi_ap_set_key(const char *key, size_t len)
+static int wifi_set_key(id_t id, const char *key, size_t len)
 {
+	key = trim(key, &len);
```
ziemleszcz: Input is already trimmed.

wi-fi/lwip/main.c
```diff
@@ -140,315 +220,863 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+		thread = wifi_sta_thread;
+		threadName = "wifi-sta";
+		other = &wifi_common.ap;
+	}
+	else {
+		return -1;
+	}
+
+#if !WIFI_CONCURRENT_MODE_ENABLED
+	if (other->tid != 0) {
```
ziemleszcz: `tid` check should be done with `dev->lock` locked.

wi-fi/lwip/main.c
```diff
@@ -140,315 +220,863 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+	keyLen = wifi_common.sta.key.len;
+
+	mutexUnlock(wifi_common.sta.lock);
+
+	left = STA_SCAN_TIMEOUT_SECONDS;
+
+	do {
+		enum wifi_state state = wifi_sta_state(&connectedTo);
+		if (state == wifi_running) {
+			if (memcmp(&connectedTo, &ssid, sizeof(ssid)) == 0) {
```
ziemleszcz: You should compare `ssid.value` using `ssid.length`. Additionally, verify that the lengths are equal, since one SSID could be a prefix of another.

wi-fi/lwip/main.c
```diff
@@ -140,315 +220,863 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+		case WHD_SCAN_ABORTED:
+			return;
+
+		case WHD_SCAN_COMPLETED_SUCCESSFULLY:
+			break;
+
+		case WHD_SCAN_INCOMPLETE: {
+			if ((*result_ptr)->SSID.length != 0) {
+				whd_ssid_t *ssid = user_data;
+				if (memcmp((*result_ptr)->SSID.value, ssid->value, ssid->length) != 0) {
```
ziemleszcz: Verify that the lengths are equal, since one SSID could be a prefix of another.

wi-fi/lwip/main.c
```diff
@@ -140,315 +220,863 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+	memcpy(key, wifi_common.sta.key.value, sizeof(key));
+	keyLen = wifi_common.sta.key.len;
+
+	mutexUnlock(wifi_common.sta.lock);
+
+	left = STA_SCAN_TIMEOUT_SECONDS;
+
+	do {
+		enum wifi_state state = wifi_sta_state(&connectedTo);
+		if (state == wifi_running) {
```
ziemleszcz: Why would STA be running here?

wi-fi/lwip/main.c
```diff
@@ -140,315 +220,863 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+			if (memcmp(&connectedTo, &ssid, sizeof(ssid)) == 0) {
+				wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_INFO, "already connected to this network\n");
+				result = WHD_SUCCESS;
+				break;
```
ziemleszcz: Result is success but ` wifiJoined` and `ifaceAdded` remain false.

wi-fi/lwip/main.c
```diff
@@ -140,315 +220,863 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+
+
+static void wifi_sta_thread(void *arg)
+{
+	(void)arg;
+	whd_result_t result;
+	whd_ssid_t ssid, connectedTo;
+	uint8_t key[WSEC_MAX_PSK_LEN];
+	uint8_t keyLen;
+	time_t when, now, left;
```
ziemleszcz: Declare `when` and `now` only if  `STA_SCAN_TIMEOUT_SECONDS != 0`?

wi-fi/lwip/main.c
```diff
@@ -140,391 +221,937 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+	mutexLock(wifi_common.ap.lock);
+	wifi_common.ap.flags |= started ? WIFI_FLAG_STARTED : WIFI_FLAG_FAILED;
+	condSignal(wifi_common.ap.cond);
+	mutexUnlock(wifi_common.ap.lock);
+
+	if (started) {
+		wifi_main_loop(&wifi_common.ap);
+	}
+
+	whd_wifi_stop_ap(wifi_common.ap.iface.whd_iface);
```
ziemleszcz: The AP is being stopped unconditionally, even if it may not have been started.

wi-fi/lwip/main.c
```diff
@@ -140,391 +221,937 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+			}
+			else if (size == CONST_STRLEN("off") && strncmp("off", data, size) == 0) {
+				mutexLock(wifi_common.lock);
+				if (!wifi_common.initialized) {
+					wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "Wi-Fi is already off\n");
+					mutexUnlock(wifi_common.lock);
+					return size;
+				}
+
+				wifi_deinit_interfaces();
```
ziemleszcz: `_wifi_deinit()` already calls `wifi_deinit_interfaces()`.

wi-fi/lwip/main.c
```diff
@@ -140,391 +221,937 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+	wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_INFO, "AP: stopped\n");
+
+	mutexLock(wifi_common.ap.lock);
+	wifi_common.ap.tid = 0;
+	if ((wifi_common.ap.flags & WIFI_FLAG_BUS_FAILED) != 0) {
+		mutexUnlock(wifi_common.ap.lock);
+
+		mutexLock(wifi_common.lock);
+		/* safe because we already set our thread's tid to 0 */
+		_wifi_deinit();
```
ziemleszcz: Why are you removing AP and STA devices? This is equivalent to running `off` command. Also note that `wifi_common.initialized` is not set to `false`.

wi-fi/lwip/main.c
```diff
@@ -140,391 +221,937 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+	wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_INFO, "STA: disconnected\n");
+
+	mutexLock(wifi_common.sta.lock);
+	wifi_common.sta.tid = 0;
+	if ((wifi_common.sta.flags & WIFI_FLAG_BUS_FAILED) != 0) {
+		mutexUnlock(wifi_common.sta.lock);
+
+		mutexLock(wifi_common.lock);
+		/* safe because we already set our thread's tid to 0 */
+		_wifi_deinit();
```
ziemleszcz: Why are you removing AP and STA devices? This is equivalent to running `off` command. Also note that `wifi_common.initialized` is not set to `false`.

wi-fi/lwip/main.c
```diff
@@ -140,391 +221,937 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+	mutexLock(wifi_common.dev[id].lock);
 
-	SNPRINTF_APPEND("running=%u\n", wifi_common.tid != 0);
-	SNPRINTF_APPEND("timeout=%u\n", wifi_common.idle_timeout);
-	SNPRINTF_APPEND("ssid=%.*s\n", wifi_common.ssid.length, wifi_common.ssid.value);
+	SNPRINTF_APPEND(overflow, buf, size, "timeout=%u\n", wifi_common.dev[id].timeout);
+	if (id == STA_DEV_ID) {
+		SNPRINTF_APPEND(overflow, buf, size, "ssid=%.*s\n", wifi_common.sta.ssid.length, wifi_common.sta.ssid.value);
+		whd_ssid_t ssid;
+		enum wifi_state state = wifi_sta_state(&ssid);
```
ziemleszcz: There is no `(dev->iface.whd_iface == NULL || whd_wifi_get_mac_address(dev->iface.whd_iface, &mac) == WHD_BUS_FAIL)` check as in `wifi_get_state()`.

wi-fi/lwip/main.c
```diff
@@ -140,391 +233,990 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
 
 	return size;
 }
 
 
-static int wifi_dev_init(unsigned int port)
+static void _wifi_ctrl_action(enum wifi_action action)
 {
-	oid_t wifi_oid = { port, WIFI_DEV_ID };
+	wifi_common.ctrl.action = action;
```
ziemleszcz: There is no protection against overwriting an ongoing action. Consider waiting for the previous action to complete before setting a new one.

julianuziemblo: Performing and setting an action is under a mutex, shouldn't that suffice?

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -687,7 +694,7 @@ whd_result_t whd_bus_usb_read_frame(whd_driver_t whd_driver, whd_buffer_t* buffe
         return whd_bus_usb_rx_queue_dequeue(buffer);
     }
 
-    return 1;
+    return WHD_BUS_FAIL;
```
ziemleszcz: Why not `WHD_NO_PACKET_TO_RECEIVE`?

wi-fi/whd/bus_protocols/whd_bus_usb_protocol.c
```diff
@@ -603,6 +606,10 @@ static whd_bool_t whd_bus_usb_wake_interrupt_present(whd_driver_t whd_driver)
  **************************************************************************************************/
 static uint32_t whd_bus_usb_packet_available_to_read(whd_driver_t whd_driver)
 {
+    if (whd_bus_usb_common.rx.failed) {
+        return WHD_BUS_FAIL;
```
ziemleszcz, line 635: It seems that this function doesn't return an error code but number of RX packets.

wi-fi/lwip/main.c
ziemleszcz: I'm not sure this is safe. What if two threads attempt to set the action concurrently? The first thread may set it, but could the second thread acquire the mutex before `condWait()` in `wifi_ctrl_thread()`?

julianuziemblo: But the ctrl thread that performs the action is under the ctrl mutex, and only releases it after handling the action - in condWait. I'm not sure how other threads could acquire the mutex. Is it possible for another than ctrl thread to be scheduled when we condSignal?

ziemleszcz: `Is it possible for another than ctrl thread to be scheduled when we condSignal?`: that's exactly what I'm worried about. `condWait` releases and reacquires the mutex. Anyway this probably implementation specific.

julianuziemblo: the only solution I see is making it atomic and waiting until the operation is complete - do you have better ideas or do you think this one will suffice?

wi-fi/lwip/main.c
```diff
@@ -140,391 +233,990 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+{
+	return whd_wifi_is_ready_to_transceive(wifi_common.sta.iface.whd_iface) != WHD_SUCCESS ? wifi_idle : wifi_running;
 }
 
 
-static void wifi_ap_main_loop(void)
+static enum wifi_state wifi_get_state(struct wifi_device *dev)
+{
+	if (dev->iface.whd_iface == NULL) {
+		return wifi_idle;
```
ziemleszcz: This should be a different state (`wifi_off`?). Originally, there was only `wifi_ap_is_idle()` function, which returned true if no client was connected to the AP and false otherwise. So, `wifi_idle` state should not be equivalent to having no AP at all.
I'm not sure how this relates to STA mode. For AP mode, we shut it down if it remains idle for `timeout` seconds. The idea was that clients can connect to DATARO AP only for a limited time after it starts (security requirement).
Do we need a similar timeout for STA mode? If so, what is its intended purpose?

wi-fi/lwip/main.c
```diff
@@ -140,391 +233,990 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
 
-		if (finish)
+		if (finish) {
 			break;
+		}
 
-		mutexLock(wifi_common.lock);
-		condWait(wifi_common.cond, wifi_common.lock, cond_timeout * 1000000ULL);
+		mutexLock(dev->lock);
+		condWait(dev->cond, dev->lock, condTimeout * 1000000ULL);
```
ziemleszcz: This appears to break AP `timeout` feature. `condTimeout` is 0 when the loop starts, resulting in an indefinite wait that is only terminated by `stop` command.
As a result, if AP `timeout` is set after the loop has started, it will never take effect.

wi-fi/lwip/main.c
```diff
@@ -140,391 +233,990 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+		wifi_dev_start(&wifi_common.sta);
+	}
+	else if (id == STA_DEV_ID && size == CONST_STRLEN("disconnect") && strncmp("disconnect", data, size) == 0) {
+		wifi_dev_stop(&wifi_common.sta);
+	}
+	else if (id == AP_DEV_ID && size == CONST_STRLEN("start") && strncmp("start", data, size) == 0) {
+		unsigned int retries = AP_START_RETRIES;
 
-		while (wifi_ap_start() < 0 && retries-- > 0) {
+		while (wifi_dev_start(&wifi_common.ap) < 0 && retries-- > 0) {
```
ziemleszcz: Probably no longer needed after last changes by jmaksymowicz.

wi-fi/lwip/main.c
ziemleszcz: Crazy idea: convert `wifi_ctrl_thread` into a full message handler (with a separate port for `/dev/wifi/ctrl`) and handle actions synchronously using the message API.
Alternatively, you could wait for the action to complete, similar to `wifi_dev_start()`, using a condition variable.
Handling `start` command synchronously would be preferable from my perspective.

julianuziemblo: hm, sounds reasonable - will give it a try

wi-fi/lwip/main.c
julianuziemblo: yes - and that's how it was in the original code. The logic was not changed. Would you like it to work the way you describe it?

ziemleszcz: OK, this is fine. It might be worth adding a comment that the timeout must be set before starting the AP.

wi-fi/lwip/main.c
julianuziemblo: The thread's purpose for STA is to check if we're still connected (e.g. the network seized to exist). You might be right it shouldn't then be able to be configured to have an infinite timeout (0). I'll try to revise it for STA.

wi-fi/lwip/main.c
```diff
@@ -140,369 +229,930 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "Wi-Fi is not up\n");
 		return;
 	}
 
-	if (portCreate(&port) < 0) {
-		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't create Wi-Fi port\n");
+	if (wifi_common.msgport != -1) {
+		portDestroy(wifi_common.msgport);
+		/* only wait for thread if we can signalize for it to close with portDestroy */
+		if (wifi_common.msgtid != 0) {
```
ziemleszcz, line 951: `msgtid` is not set to the actual `TID`.

wi-fi/lwip/main.c
```diff
@@ -140,369 +229,930 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+		/* keep this as the last thing here so that we don't have to deinitialize it on error */
+		err = sys_thread_new("wifi-msg", wifi_msg_thread, NULL, WIFI_THREAD_STACKSZ, WIFI_THREAD_PRIO);
+		if (err < 0) {
+			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't create Wi-Fi msgthread\n");
+			break;
+		}
+	} while (0);
+
+	/* set this regardless as it will be set to false in _wifi_deinit if initialization failed */
+	wifi_common.initialized = true;
```
ziemleszcz: Unsafe, `wifi_msg_thread` could try to handle some message when initialization failed.

wi-fi/lwip/main.c
```diff
@@ -140,369 +229,930 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+static void _wifi_deinit(void)
+{
+	if (!wifi_common.initialized) {
+		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "Wi-Fi is not up\n");
 		return;
 	}
 
-	if (portCreate(&port) < 0) {
-		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't create Wi-Fi port\n");
+	if (wifi_common.msgport != -1) {
```
ziemleszcz: `msgport` is unsigned.

wi-fi/lwip/main.c
```diff
@@ -140,369 +229,930 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+static int _wifi_init(void)
+{
+	if (wifi_common.initialized) {
+		wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "Wi-Fi is already initialized\n");
+		return 0;
+	}
+
+	int err;
+
+	wifi_common.msgport = -1;
```
ziemleszcz: `msgport` is unsigned.

wi-fi/lwip/main.c
ziemleszcz, line 951: `msgport` is not set to 0 if `wifi_common.msgtid == 0` (created msg thread failed).

wi-fi/lwip/main.c
```diff
@@ -140,369 +229,930 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+		portDestroy(wifi_common.msgport);
+		/* only wait for thread if we can signalize for it to close with portDestroy */
+		if (wifi_common.msgtid != 0) {
+			sys_thread_join(wifi_common.msgtid);
+			wifi_common.msgtid = 0;
+		}
+	}
+
+	wifi_deinit_interfaces();
+	if (wifi_common.sta.lock != -1) {
```
ziemleszcz, line 958: What if somebody reads `/dev/wifi/ap` or `/dev/wifi/sta` at this point? These devices are destroyed at the end of this function.

julianuziemblo: We've already joined the msgthread by then, so that's not possible

julianuziemblo: Should we first delete the devices and then join the msgthread?

wi-fi/lwip/main.c
```diff
@@ -140,369 +229,944 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+{
+	data = trim(data, &size);
+
+	if (size >= CONST_STRLEN("ssid ") && strncmp("ssid ", data, 5) == 0) {
+		wifi_set_ssid(id, data + 5, size - 5);
+	}
+	else if (size >= CONST_STRLEN("key ") && strncmp("key ", data, 4) == 0) {
+		wifi_set_key(id, data + 4, size - 4);
+	}
+	else if (id == STA_DEV_ID && size == CONST_STRLEN("connect") && strncmp("connect", data, size) == 0) {
```
ziemleszcz, line 758: I'm not sure about a synchronous STA connection. How long could it block if no network is available? This operation might be better handled asynchronously.

wi-fi/lwip/main.c
ziemleszcz: OK, you are right.

wi-fi/lwip/main.c
julianuziemblo: It will block no more than STA_TIMEOUT seconds defined at the top of the file

julianuziemblo: ... But I guess I made it configurable, so maybe it should be async

ziemleszcz: OK, I'll leave the decision to you. For 10 seconds, it could remain synchronous. My main concern was preserving synchronous AP operations.

wi-fi/lwip/main.c
```diff
@@ -140,369 +238,944 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+		err = sys_thread_opt_new("wifi-msg", wifi_msg_thread, NULL, WIFI_THREAD_STACKSZ, WIFI_THREAD_PRIO, &wifi_common.msgtid);
+		if (err < 0) {
+			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't create Wi-Fi msgthread\n");
+			break;
+		}
+	} while (0);
+
+	if (err < 0) {
+		_wifi_deinit();
+	}
```
ziemleszcz, line 1060: `else wifi_common.initialized = true`?

wi-fi/lwip/main.c
```diff
@@ -140,369 +238,944 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+	wifi_common.ap.iface.whd_iface = NULL;
+}
+
+
+static void _wifi_deinit(void)
+{
+	destroy_dev(AP_DEV_NAME);
+	destroy_dev(STA_DEV_NAME);
+
+	if (wifi_common.msgport != (unsigned int)-1) {
```
ziemleszcz: Nitpick: `(uint32_t)-1`.

wi-fi/lwip/main.c
```diff
@@ -140,369 +238,944 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+		portDestroy(wifi_common.msgport);
+		wifi_common.msgport = (unsigned int)-1;
```
ziemleszcz: Nitpick: `(uint32_t)-1`.

wi-fi/lwip/main.c
```diff
@@ -140,369 +238,944 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+	}
+	mutexUnlock(wifi_common.ctrl.lock);
+}
+
+
+static int _wifi_init(void)
+{
+	int err;
+
+	wifi_common.msgport = (unsigned int)-1;
```
ziemleszcz: Nitpick: `(uint32_t)-1`.

wi-fi/lwip/main.c
```diff
@@ -140,369 +238,944 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+				}
+				else if (size == CONST_STRLEN("off") && strncmp("off", data, size) == 0) {
+					wifi_deinit();
+				}
+				else {
+					wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "got unknown Wi-Fi command: %.*s\n", (int)size, data);
+					msg.o.err = -EINVAL;
+					break;
+				}
+				msg.o.err = size;
```
ziemleszcz: ` = msg.i.size`?

wi-fi/lwip/main.c
```diff
@@ -140,369 +238,944 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
-				msg.o.err = wifi_dev_write(msg.i.data, msg.i.size);
+			case mtWrite: {
+				const char *data = msg.i.data;
+				size_t size = msg.i.size;
+
+				data = trim(data, &size);
+
+				if (size == CONST_STRLEN("on") && strncmp("on", data, size) == 0) {
+					if (wifi_init() < 0) {
+						wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "failed to initialize Wi-Fi\n");
```
ziemleszcz: Not true if Wi-Fi is already intialized. Maybe move this log to `wifi_init()` or return 0 in `wifi_init()` for this case.

wi-fi/lwip/main.c
```diff
@@ -140,369 +238,944 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+	else if (id == AP_DEV_ID && size >= CONST_STRLEN("timeout ") && strncmp("timeout ", data, 8) == 0) {
+		wifi_ap_set_timeout(data + 8, size - 8);
+	}
+	else if (id == AP_DEV_ID && size == CONST_STRLEN("start") && strncmp("start", data, size) == 0) {
+		wifi_ap_start();
+	}
+	else if (id == AP_DEV_ID && size == CONST_STRLEN("stop") && strncmp("stop", data, size) == 0) {
 		wifi_ap_stop();
 	}
+	else if (id == AP_DEV_ID && size >= CONST_STRLEN("ipaddr ") && strncmp("ipaddr ", data, 8) == 0) {
```
ziemleszcz: 8 -> 7

wi-fi/lwip/main.c
julianuziemblo: technically it is true then :smile: but I see what you mean, will move it

wi-fi/lwip/main.c
```diff
@@ -140,369 +238,949 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+		result = whd_wifi_start_ap(wifi_common.ap.iface.whd_iface);
+		if (result != WHD_SUCCESS) {
+			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't start Wi-Fi AP\n");
+			break;
+		}
 
-		result = cybsp_init();
-		if (result != CY_RSLT_SUCCESS) {
-			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't init Wi-Fi HW\n");
+		if (wifi_add_interface(&wifi_common.ap.iface, &wifi_common.ap.apPriv.addr) < 0) {
```
ziemleszcz: Missing `whd_wifi_stop_ap()` call on failure.

wi-fi/lwip/main.c
```diff
@@ -140,369 +238,949 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+	mutexLock(wifi_common.dev[id].lock);
+
+	if (id == STA_DEV_ID) {
+		SNPRINTF_APPEND(overflow, buf, size, "ssid=%.*s\n", wifi_common.sta.ssid.length, wifi_common.sta.ssid.value);
+		enum wifi_state state = wifi_get_state(&wifi_common.sta);
+		if (state == wifi_running) {
+			whd_ssid_t ssid;
+			(void)wifi_sta_get_connected(&ssid);
+			struct netif *netif = cy_lwip_get_interface(wifi_common.dev[id].iface.role);
+			SNPRINTF_APPEND(overflow, buf, size, "current=%.*s\n", ssid.length, ssid.value);
```
ziemleszcz: `ssid` is uninitialized on `wifi_sta_get_connected()` failure.

wi-fi/lwip/main.c
```diff
@@ -140,369 +238,949 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+			}
+			result = whd_wifi_scan(wifi_common.sta.iface.whd_iface, STA_SCAN_TYPE, STA_BSS_TYPE,
+					&ssid, NULL, NULL, NULL, wifi_scan_result_cb, &wifi_common.sta.staPriv.scanResult, &ssid);
+			condWait(wifi_common.sta.cond, wifi_common.sta.lock, min((WIFI_STA_SCAN_TIMEOUT_SECONDS * 1000000ULL) / WIFI_STA_N_SCANS, left * 1000000ULL));
+
+			(void)whd_wifi_stop_scan(wifi_common.sta.iface.whd_iface);
+
+			if (clock_gettime(CLOCK_MONOTONIC, &tp) < 0) {
+				wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "couldn't get current time: %s (%d)\n", strerror(errno), errno);
+				break;
```
ziemleszcz, line 513: result = WHD_SCAN_ABORTED

wi-fi/lwip/main.c
```diff
@@ -140,369 +238,958 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
-			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't init Wi-Fi HW\n");
+		if (wifi_add_interface(&wifi_common.ap.iface, &wifi_common.ap.apPriv.addr) < 0) {
+			result = WHD_INTERFACE_NOT_UP;
 			break;
 		}
+		ifaceAdded = true;
+	} while (0);
+
+	mutexLock(wifi_common.ap.lock);
+	wifi_common.ap.apPriv.flags |= started ? WIFI_FLAG_STARTED : WIFI_FLAG_FAILED;
```
ziemleszcz: started -> result == WHD_SUCCESS

wi-fi/lwip/main.c
```diff
@@ -140,369 +238,958 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+		if (result != WHD_SUCCESS) {
+			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't start Wi-Fi AP\n");
+			break;
+		}
+		started = true;
 
-		result = cybsp_init();
-		if (result != CY_RSLT_SUCCESS) {
-			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "can't init Wi-Fi HW\n");
+		if (wifi_add_interface(&wifi_common.ap.iface, &wifi_common.ap.apPriv.addr) < 0) {
```
ziemleszcz: Suggestion: `wifi_add_interface()` could return `whd_result_t` directly.

wi-fi/lwip/main.c
```diff
@@ -140,369 +238,957 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
-	else if (strncmp("start", data, size) == 0) {
-		unsigned int retries = WIFI_START_RETRIES;
+}
 
-		while (wifi_ap_start() < 0 && retries-- > 0) {
-			/* FIXME: temporary workaround - find out why AP doesn't start */
-			wm_cy_log_msg(CYLF_MIDDLEWARE, CY_LOG_ERR, "retrying to start Wi-Fi AP\n");
-		}
+
+static int wifi_dev_write(id_t id, const char *data, size_t size)
```
ziemleszcz, line 747: This function should return back exactly `size` on success (not trimmed size).

wi-fi/lwip/main.c
```diff
@@ -140,369 +238,957 @@ static int wifi_ap_set_key(const char *key, size_t len)
...
+			SNPRINTF_APPEND(overflow, buf, size, "ipaddr=%s\n", netif != NULL ? ipaddr_ntoa(&netif->ip_addr) : "unknown");
+			SNPRINTF_APPEND(overflow, buf, size, "netmask=%s\n", netif != NULL ? ipaddr_ntoa(&netif->netmask) : "unknown");
+			SNPRINTF_APPEND(overflow, buf, size, "gateway=%s\n", netif != NULL ? ipaddr_ntoa(&netif->gw) : "unknown");
+		}
+		SNPRINTF_APPEND(overflow, buf, size, "running=%u\n", state == wifi_running ? 1U : 0U);
+	}
+	else if (id == AP_DEV_ID) {
+		SNPRINTF_APPEND(overflow, buf, size, "timeout=%u\n", wifi_common.ap.apPriv.timeout);
+		SNPRINTF_APPEND(overflow, buf, size, "ssid=%.*s\n", wifi_common.ap.ssid.length, wifi_common.ap.ssid.value);
+		SNPRINTF_APPEND(overflow, buf, size, "ip_addr=%s\n", ipaddr_ntoa(&wifi_common.ap.apPriv.addr.addr));
```
ziemleszcz: `ipaddr` is used for STA.

ziemleszcz (dismissed):

ziemleszcz (approved):

PR #144, opened 2026-06-02
Changes to enable PPPOS support on imx6ull 

anglov:
## Description

Some cleanups, by renaming, removing configuration from modem, change prefix of defines and sensible warning/error on misconfiguration and change argument parsing

## Motivation and Context

Changes for using pppos modem on imx6ull target, especially SEN

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

- Changing PPPOS_DEFAULT_APN define prefix breaks BES configuration, of modem parameters in build time. Any project overwriting this  in build time should modify PPPOS_DEFAULT_APN to PPPOS_MODEM_DEFAULT_APN
- Arguments for pppos need to be tuned to match current specification
- Projects on imx6ull that doesn't want to build pppo[s/u] driver, should configure NET_DRIVERS accordingly.
 
## How Has This Been Tested?

- [x] Tested by hand on: SEN

## Checklist:

## Special treatment
- [x] This PR needs additional PRs to work (list the PRs, preferably in merge-order).
https://github.com/phoenix-rtos/phoenix-rtos-project/pull/1651
- [x] I will merge this PR by myself when appropriate.

### DIFF
--- a/drivers/pppos.c
+++ b/drivers/pppos.c
@@ -679,8 +679,13 @@ fail:
 static int pppos_netifUp(pppos_priv_t *state)
 {
 #if PPPOS_USE_CONFIG_FILE
+	if (state->config_path == NULL) {
+		return 1;
+	}
+
 	char lcfg[256] = { 0 };
 	int line = 0;
+
 	FILE *fcfg = fopen(state->config_path, "r");
 	char *cfgval;
 	char *eq;
@@ -774,27 +779,59 @@ static void pppos_statusCallback(struct netif *netif)
 }
 
 
-static char *cfg_get_next_arg(char *arg)
-{
-	if (arg == NULL || *arg == '\0')
-		return NULL;
+struct iter {
+	char *next;
+};
 
-	for (; *arg; arg++) {
-		if (*arg == ':') {
-			*arg++ = '\0';
-			break;
-		}
+
+static char *iterateArgs(struct iter *iter)
+{
+	if (iter->next == NULL) {
+		return NULL;
 	}
 
-	return arg;
+	char *const cur = iter->next;
+	char *curEnd = strchr(cur, ':');
+	if (curEnd != NULL) {
+		*curEnd = '\0';
+		iter->next = curEnd + 1;
+	}
+	else {
+		iter->next = NULL;
+	}
+
+	return cur;
 }
 
 
+static void initIter(struct iter *iter, char *arg)
+{
+	iter->next = arg;
+}
+
+/*
+ * Arguments:
+ * <control_device>[:<option>...]
+ *
+ * <control_device> Path to the primary control device (starts with AT, switches into PPP) (required)
+ * <option>         Zero or more optional option separated by colons (':')
+ *
+ * Supported options:
+ * st=<status_device> Status device, for retrieving media information (additional AT interface)
+ * cfg=<config_file>  Path to config file, when PPPOS_USE_CONFIG_FILE enabled
+ * up                 Bring the interface up on start
+ * nodefault          Disable default route
+ * nodns              Disable DNS
+ *
+ * Examples:
+ * /dev/usbacm0
+ * /dev/usbacm0:up:st=/dev/usbacm2
+ * /dev/usbacm0:up:nodefault:nodns:st=/dev/usbacm2
+ */
 static int pppos_netifInit(struct netif *netif, char *cfg)
 {
-	pppos_priv_t* state;
+	pppos_priv_t *state;
 	int retries, flags = 0;
-	char *next;
 
 	// NOTE: netif->state cannot be used to keep our private state as it is used by LWiP PPP implementation, pass it as *ctx to callbacks
 	state = netif->state;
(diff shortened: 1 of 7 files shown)


modem/telit/pppos_modem.h
```diff
@@ -14,15 +14,14 @@
...
 #include <stddef.h>
 
-#define PPPOS_USE_CONFIG_FILE     1
-#define PPPOS_USE_AUTH            0
-#define PPPOS_DISCONNECT_ON_INIT  1
-#define AT_CONNECT_CMD            "AT+CGDATA=\"PPP\",1\r\n"
-#define AT_DISCONNECT_CMD         "ATH\r\n"
-#define AT_INIT_CMDS_TIMEOUT_MS   3000
-#define AT_CONNECT_CMD_TIMEOUT_MS 3000
+#define PPPOS_MODEM_USE_AUTH                  0
```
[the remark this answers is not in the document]
anglov: Please read commit message. This not depending on modem, rather on platform. As this modem was used on legacy INC platform, no further moves required.

drivers/pppos.c
```diff
@@ -28,6 +34,33 @@
 
 #include <pppos_modem.h>
 
+
+#if PPPOS_MODEM_USE_AUTH == 1 && PPP_AUTH_SUPPORT == 0
+#error Modem configured for authentication, but no method is enabled!
+#endif
```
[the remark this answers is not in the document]
anglov: ... and code above exactly shows why I don't want to apply that checks. LWIP upstream just causes runtime failure in that  case, so I think we may too.

In desired solution, we shouldn't hardcode modem config.

drivers/pppos.c
```diff
@@ -756,77 +784,118 @@ static void pppos_statusCallback(struct netif *netif)
...
 
-	for (; (next = cfg_get_next_arg(cfg)); cfg = next) {
-		if (!strncmp(cfg, "/dev/", 5)) {
-			state->serialat_fn = cfg;
-			log_info("config device: %s", cfg);
-			continue;
-		}
+	if (state->serialdev_fn == NULL) {
+		return ERR_ARG;
+	}
```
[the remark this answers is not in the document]
anglov: @gemini-code-assist – passing NULL to open is undefined behavior. Passing empty string is well-defined behavior.  If you think deeper, then you may think, that passing "/" is also wrong. And then we came to conclusion that you need to do full path validating, which is not possible at this stage (file may appear after run)

drivers/pppos.c
```diff
@@ -756,77 +784,118 @@ static void pppos_statusCallback(struct netif *netif)
...
+		else if (strcmp(arg, "nodns") == 0) {
 			flags |= CFG_FLAG_NO_DNS;
 			log_info("config no DNS: yes");
-			continue;
 		}
+		else {
+			log_error("Not recognized argument: %s", arg);
+			return ERR_ARG;
+		}
+	}
```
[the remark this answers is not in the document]
anglov: @gemini-code-assist – passing NULL to open is undefined behavior. Passing empty string is well-defined behavior.  If you think deeper, then you may think, that passing "/" is also wrong. And then we came to conclusion that you need to do full path validating, which is not possible at this stage (file may appear after run)

drivers/pppos.c
```diff
@@ -661,8 +679,18 @@ static void pppos_mainLoop(void* _state)
 static int pppos_netifUp(pppos_priv_t *state)
 {
 #if PPPOS_USE_CONFIG_FILE
+	/*
+	 * FIXME: This code require verification before production use. Spotted issues:
+	 * - early success return, without setting want_connected flag and signaling cond
+	 * - waiting for state->apn[0] in mainloop which may be never set here
+	 */
```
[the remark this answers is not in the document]
anglov: @gemini-code-assist – that's why I put FIXME.

drivers/pppos.c
```diff
@@ -35,6 +35,11 @@
 #include <pppos_modem.h>
 
 
+#if PPPOS_MODEM_USE_AUTH == 1 && PPP_AUTH_SUPPORT == 0
+#error Modem configured for authentication, but no method is enabled!
```
julianuziemblo, line 39: nitpick: add quotations around the error message

drivers/pppos.c
```diff
@@ -9,6 +9,12 @@
  * %LICENSE%
  */
 
+#include "lwip/opt.h"
+
+#if PPP_SUPPORT == 0 || PPPOS_SUPPORT == 0
+#warning pppos driver requires both PPP_SUPPORT and PPPOS_SUPPORT. Disabling pppos driver.
```
julianuziemblo, line 15: nitpick: add quotations around the warning message

drivers/pppou.c
```diff
@@ -9,6 +9,12 @@
  * %LICENSE%
  */
 
+#include "lwip/opt.h"
+
+#if PPP_SUPPORT == 0 || PPPOS_SUPPORT == 0
+#warning pppou driver requires both PPP_SUPPORT and PPPOS_SUPPORT. Disabling pppou driver.
```
julianuziemblo, line 15: nitpick: add quotations around the warning message

drivers/pppos.c
```diff
@@ -774,77 +779,118 @@ static void pppos_statusCallback(struct netif *netif)
...
-			continue;
-		}
+	if (state->serialdev_fn == NULL) {
+		return ERR_ARG;
+	}
 
-		if (strcmp(cfg, "up") == 0) {
+	for (char *arg = iterateArgs(&iter); arg != NULL; arg = iterateArgs(&iter)) {
+		if (strncmp(arg, "st=", 3) == 0) {
+			const size_t prefixLen = 3;
```
julianuziemblo: this seems redundant, just `+ 3` in the next line would be enough for me. I know that prefixLen is more descriptive, but to me this also suggests that this will be used somewhere else somehow, which it of course isn't, which makes this somewhat confusing

drivers/pppos.c
anglov: I would prefer to not to add. 

Let's assume code:
```c
#include <stdio.h>
#warning test1
#warning "test2"

int f()
{
        printf("test\n");
}

int main()
{
        f();
}
```

GCC result of compilation:
```
$ gcc -Wall -Wextra a.c 
a.c:2:2: warning: #warning test1 [-Wcpp]
    2 | #warning test1
      |  ^~~~~~~
a.c:3:2: warning: #warning "test2" [-Wcpp]
    3 | #warning "test2"
      |  ^~~~~~~
a.c: In function ‘f’:
a.c:8:1: warning: control reaches end of non-void function [-Wreturn-type]
    8 | }
      | ^
```

clang result of compilation:
```
$ clang -Wall -Wextra a.c 
a.c:2:2: warning: test1 [-W#warnings]
    2 | #warning test1
      |  ^
a.c:3:2: warning: "test2" [-W#warnings]
    3 | #warning "test2"
      |  ^
a.c:8:1: warning: non-void function does not return a value [-Wreturn-type]
    8 | }
      | ^
3 warnings generated.
```

As you can see – adding "" just produces additional "" – in GCC it's rather who cares looking at output, but compiling with clang causes worse output message.

It's also well-defined in C standard (from C23 for warning, lol, for errors from C89), that space separated identifiers is good. So if you don't have any good reason to add quotation, I would prefer to not add it :)

julianuziemblo: ok, just seemed that we use quotations everywhere else for warnings/errors. Feel free to leave at as-is though.

drivers/pppos.c
anglov: Yeaaah, remaining on other approach that I developed before.

drivers/pppos.c
```diff
@@ -844,8 +911,8 @@ static int pppos_netifInit(struct netif *netif, char *cfg)
 			ppp_set_usepeerdns(state->ppp, 1);
 #endif /* LWIP_DNS */
 
-#if PPPOS_USE_AUTH
-		ppp_set_auth(state->ppp, PPPOS_AUTH_TYPE, PPPOS_AUTH_USER, PPPOS_AUTH_PASSWD);
+#if PPPOS_MODEM_USE_AUTH && PPP_AUTH_SUPPORT
+		ppp_set_auth(state->ppp, PPPOS_MODEM_AUTH_TYPE, PPPOS_MODEM_AUTH_USER, PPPOS_MODEM_AUTH_PASSWD);
 #endif /* PPPOS_USE_AUTH */
```
ziemleszcz: Change the comment to `/* PPPOS_MODEM_USE_AUTH && PPP_AUTH_SUPPORT */` or better remove it (it's only 1 line of code).

ziemleszcz (approved):

